PR #201, opened 2017-12-17
Expose TypeInfo on ParserResult<T>

Fumblesneeze:
I ran across the problem of having to know which verb had been selected when an error occured. The `ParserResult` would be a `NotParsed<object>` because I use the ParseArguments overload that takes an array of types. From that result there was no way to obtain the type of the verb except with some reflection to access these private members, so I request that typeinfo be made available.

Code now:

    var typeinfo = result.GetType().GetProperty("TypeInfo", BindingFlags.NonPublic | BindingFlags.Instance).GetValue(result);
    var option = (Type)typeinfo.GetType().GetProperty("Current", BindingFlags.Public | BindingFlags.Instance).GetValue(typeinfo);

Code then:

    var option = result.TypeInfo.Current;

### DIFF
--- a/src/CommandLine/ParserResult.cs
+++ b/src/CommandLine/ParserResult.cs
@@ -6,7 +6,7 @@ using System.Linq;
 
 namespace CommandLine
 {
-    sealed class TypeInfo
+    public sealed class TypeInfo
     {
         private readonly Type current;
         private readonly IEnumerable<Type> choices; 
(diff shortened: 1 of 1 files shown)


ericnewton76 (commented):
I'm wondering if we should approach this from a different way, making an internal class public can have mutability consequences.

Perhaps the ParserResult class itself should have some kind of type info available that doesnt leak out all the internals

PR #205, opened 2018-01-05
Add support for TimeSpan

0xced:
This is a work in progress, please do not merge yet.

### DIFF
--- a/src/CommandLine/Core/TypeConverter.cs
+++ b/src/CommandLine/Core/TypeConverter.cs
@@ -51,9 +51,17 @@ namespace CommandLine.Core
 
         private static object ConvertString(string value, Type type, CultureInfo conversionCulture)
         {
-            return Convert.ChangeType(value, type, conversionCulture);
+            try
+            {
+                return Convert.ChangeType(value, type, conversionCulture);
+            }
+            catch (InvalidCastException)
+            {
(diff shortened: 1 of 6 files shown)


0xced:
It looks fine now. You can merge if you think this is a good addition.

0xced:
Oh and by the way, I found why 30 tests were failing: because `Console.WindowWidth` returns `0` when running under mono leading to `ArgumentOutOfRangeException` being thrown because [maximumLength < 1](https://github.com/commandlineparser/commandline/blob/f68a93f97c959697fa134222dca6344d2ebfac5f/src/CommandLine/Text/HelpText.cs#L634-L637).

nemec:
This is great - not only will it work for TimeSpans, but any class with a TypeConverter should be supported by the library now.

Thank you for contributing, @0xced!

ericnewton76:
@0xced hey cedric, regarding the tests and the maximumLength, you are exactly right.  I was trying to find some time to implement that aspect in the testing apparatus to force that variable to always be 80 instead of -1 if the Console.WindowWidth access failed.  You'll probably see a pull request from me regarding that at some point

0xced:
I noticed that this pull request was merged (into develop) but only commit 6e52805a11a455396581b7b3d8a168ec1c6e1ff3 was cherry-picked for version 2.2.1. Are you going to merge it in into master for the next release?

PR #218, opened 2018-01-12
Updated README examples

kazenshi:
Updated the C# quick start example to work with the new API.

Added the VB.NET quick start example. 

Also I added the VB.NET Verb example.

@ericnewton76


ericnewton76:
Thats awesome. Thanks!  Merging now

@nemec (since this is readme of current version, going straight to master)

Issue #224, opened 2018-01-14
How to remove empty line between each options in help display ?


sybaris:
Hello

When I have several options, the display of help show me an empty line between each option.
I saw the property **AdditionalNewLineAfterOption** in HelpText.
I would like to set this to false for AutoBuild, but how can I do that ?

Thanks for advance
Sybaris

DominikWiesend:
This does the Job for me, but I'm struggling getting a additional line between...
ERROR(S)
USAGE

If you wanna have the copyrights as well, use another constructor for [HelpText] (new HelpText(SentenceBuilder.Create(), ..., ....)


```
Parser myParser = new Parser(config => config.HelpWriter = null);
ParserResult<object> myResult = myParser.ParseVerbs<DUALHOME, MOVE>(args);

myResult.WithNotParsed(errs =>
{
    HelpText myHelpText = HelpText.AutoBuild(myResult, onError =>
    {
        HelpText nHelpText = new HelpText(SentenceBuilder.Create(), "MS Operation Manager - AgentManagement (agtmve)");
        nHelpText.AdditionalNewLineAfterOption = false;
        nHelpText.AddDashesToOption = true;
        nHelpText.MaximumDisplayWidth = 4000;
        nHelpText.AddOptions(myResult);
        return HelpText.DefaultParsingErrorsHandler(myResult, nHelpText);
    },
    onExample => { return onExample; });
    Console.Error.WriteLine(myHelpText);
});
```

sybaris:
Hi 

Thanks for your answer. 
It's works when I have an error : help is displayed as I want. But when I call my program with the --help option, I still have  an empty line between each option. I don't understand why...

Regards,
Sybaris

PR #242, opened 2018-02-12
Added a feature to avoid linebreaks between option when displaying help

sybaris:
Hi

Linked with issue #224.
I try to make this to allow users to set option in ParserSettings to have or not linebreaks between option when displaying help.

Thanks for advance

### DIFF
--- a/src/CommandLine/Parser.cs
+++ b/src/CommandLine/Parser.cs
@@ -183,7 +183,7 @@ namespace CommandLine
                     settings.EnableDashDash)(arguments, optionSpecs);
         }
 
-        private static ParserResult<T> MakeParserResult<T>(ParserResult<T> parserResult, ParserSettings settings)
+        private /*static*/ ParserResult<T> MakeParserResult<T>(ParserResult<T> parserResult, ParserSettings settings)
         {
             return DisplayHelp(
                 parserResult,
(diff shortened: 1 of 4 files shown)


ericnewton76:
In it's current state, you actually break the interface for `HelpText.AutoBuild`.  You need to preserve the existing method parameters with a default `ParserSettings` instance.

ericnewton76 (changes requested):
In it's current state, you actually break the interface for HelpText.AutoBuild. You need to preserve the existing method parameters with a default ParserSettings instance.

You can always mark the older method as Obsolete if it makes sense.

src/CommandLine/Parser.cs
```diff
@@ -183,20 +183,20 @@ public void Dispose()
                     settings.EnableDashDash)(arguments, optionSpecs);
         }
 
-        private static ParserResult<T> MakeParserResult<T>(ParserResult<T> parserResult, ParserSettings settings)
+        private /*static*/ ParserResult<T> MakeParserResult<T>(ParserResult<T> parserResult, ParserSettings settings)
```
ericnewton76, line 186: It would be better to just remove the static modifier altogether.  We can see the change in diffs.

src/CommandLine/Parser.cs
```diff
@@ -183,20 +183,20 @@ public void Dispose()
...
         {
             return DisplayHelp(
                 parserResult,
                 settings.HelpWriter,
                 settings.MaximumDisplayWidth);
         }
 
-        private static ParserResult<T> DisplayHelp<T>(ParserResult<T> parserResult, TextWriter helpWriter, int maxDisplayWidth)
+        private /*static*/ ParserResult<T> DisplayHelp<T>(ParserResult<T> parserResult, TextWriter helpWriter, int maxDisplayWidth)
```
ericnewton76, line 194: It would be better to just remove the static modifier altogether.  We can see the change in diffs.

tests/CommandLine.Tests/Unit/Text/HelpTextTests.cs
```diff
@@ -315,7 +315,7 @@ public void Invoke_AutoBuild_for_Options_returns_appropriate_formatted_text()
                     });
 
             // Exercize system
-            var helpText = HelpText.AutoBuild(fakeResult);
+            var helpText = HelpText.AutoBuild(fakeResult, new ParserSettings());
```
ericnewton76, line 318: All of these test updates should've alerted you to the fact that you broke the existing expected interface.  This will cause problems for people simply upgrading from a patch release, we'd have to make it wait until a full version release.

As noted elsewhere, you should preserve this method and make your new method an overload with the old method calling the new method with that default `ParserSettings` instance.

ericnewton76 (changes requested):

Issue #200, opened 2017-12-13
Is there a way to override the "version" command?


Eason-Lian:
I'm trying to write some custom logic in version command. Is there a way to do that? I'm using the latest version (v2.0.275.0).

ericnewton76:
Can you clarify a little more of what your doing, whats happening, and what you were expecting?

Eason-Lian:
@ericnewton76 Thanks for replying! For some reason, I'm not able to modify assembly information. That means the version which "--version" shows is not correct. 

What I want is to specify assembly version in a configuration file (xml/json) and then read them when user is typing "--version". I created a custom VersionOptions in my project but unfortunately it can't be reached. Is there a way to override the default version option?

stabiliser:
I have the same problem, am wanting to have a 'version' command to retrieve the version from a server, but CommandLineParser is always outputting the client version if I use 'version'.

Issue #87, opened 2017-11-04
Possibility to turn off or customize version/help option


ericnewton76:
<a href="https://github.com/ptylenda"><img src="https://avatars3.githubusercontent.com/u/11139004?v=4" align="left" width="96" height="96" hspace="10"></img></a> **Issue by [ptylenda](https://github.com/ptylenda)**
_Monday Sep 19, 2016 at 14:43 GMT_
_Originally opened as https://github.com/gsscoder/commandline/issues/351_

----

It would be a good idea to make it possible to turn off (or at least customize) version and help options, probably via an attribute for the options class. This especially concerns version option - we are migrating from old custom parsers to the CommandLine library and we have noticed that it conflicts with our custom version option (-v, --version). This option is unrelated to version of an application, it concerns version of a file being processed by the executable. In our case it is not possible to use a different option name as we have to maintain 100% backward compatibility inside an enterprise system. Fortunately this "conflict" seems to be only visible in the help screen, when both our custom option and the default option are displayed at once.

ericnewton76:
<a href="https://github.com/nemec"><img src="https://avatars3.githubusercontent.com/u/136378?v=4" align="left" width="48" height="48" hspace="10"></img></a> **Comment by [nemec](https://github.com/nemec)**
_Saturday Nov 05, 2016 at 07:22 GMT_

----

This is a good idea. Right now both `--help` and `--version` are hardcoded into the application.

ericnewton76:
<a href="https://github.com/cjbush"><img src="https://avatars2.githubusercontent.com/u/5855938?v=4" align="left" width="48" height="48" hspace="10"></img></a> **Comment by [cjbush](https://github.com/cjbush)**
_Tuesday Jan 24, 2017 at 00:23 GMT_

----

+1

PR #256, opened 2018-03-20
Added AutoHelp and AutoVersion properties to control adding of implicit 'help' and 'version' options/verbs

Athari:
Covers #87 and #200.

### DIFF
--- a/src/CommandLine/Core/InstanceBuilder.cs
+++ b/src/CommandLine/Core/InstanceBuilder.cs
@@ -23,6 +23,8 @@ namespace CommandLine.Core
             StringComparer nameComparer,
             bool ignoreValueCase,
             CultureInfo parsingCulture,
+            bool autoHelp,
+            bool autoVersion,
             IEnumerable<ErrorType> nonFatalErrors)
         {
             var typeInfo = factory.MapValueOrDefault(f => f().GetType(), typeof(T));
@@ -129,7 +131,7 @@ namespace CommandLine.Core
             };
 
             var preprocessorErrors = arguments.Any()
-                ? arguments.Preprocess(PreprocessorGuards.Lookup(nameComparer))
+                ? arguments.Preprocess(PreprocessorGuards.Lookup(nameComparer, autoHelp, autoVersion))
                 : Enumerable.Empty<Error>();
 
             var result = arguments.Any()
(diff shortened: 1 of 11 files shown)


nemec:
I think this solves the issues nicely. Can you add a few tests to help us prevent regressions? I think the following should cover things, but if you think of any other scenarios feel free to include them:

1. AutoHelp = false and invoking `--help` returns an error.
2. AutoVersion = false and invoking `--version` returns an error.
3. AutoHelp = false, a flag for `--help` added manually by the user to the Options class, and it can be invoked successfully.
4. AutoVersion = false, a flag for `--version` added manually by the user to the Options class, and it can be invoked successfully.

Athari:
@nemec Added tests.

brphelps:
@nemec Can it be merge time?

moh-hassan:
@Athari
Please, can you add a wiki documentation page to explain how to use the new feature of AutoHelp/ AutoVersion with Code examples.
There is a confusion and misuse of this feature.

moh-hassan:
@athari
Please can you confirm if this feature 
[ ] can enable developer to provide his own help/ version text.
[ ] show/ hide help/version.

Why setting  exist in  both parserSetting and help text.
There is a misunderstanding in using  this feature.

CC / @rmunn

Athari:
@moh-hassan 
Looks like I've messed up something. I did have problems propagating options, but instead of introducing heavier changes just added two more arguments here and there. But the main reason for me missing the bugs is likely correctly identified by @x789:

> #256 describes the possibility to automatically add a 'help' and 'version' verb/option to the parser. This works as far as only the parser is considered. I think when #256 was integrated, it was not considered that there will also be applications without 'help' and 'version' (see #294, #414 and #455).

What others have commented on the #461 is correct: I'd expect `AutoHelp=false` and `AutoVersion=false` to allow both providing own commands and not providing any (see [comment above](https://github.com/commandlineparser/commandline/pull/256#issuecomment-376300704)).

Anyway, what's the current status? Wasn't #461 closed?

P.S. Frankly, I consider this library a lost cause, mostly because the original developer went crazy with the pseudo-functional style and made a mess of both the public API (tons of callbacks) and the implementation ("maybe" and other reinvented wheels with very poor implementations), while completely ignoring good old code smells like methods with tons of arguments, and I doubt anybody is going to spend time refactoring the whole codebase and making massive backwards compatbility breaking changes to public API yet again, even if it's allowed by the current maintainers.

henon:
Just giving a heads up. It is possible to override --help and --version now if you set AutoHelp and AutoVersion to false in the parser config.

Issue #227, opened 2018-01-21
Version 2 pulls in many NuGet dependencies


RobSmyth:
Hi,

On upgrade from 1.9 to the latest ver 2 I found that there were many NuGet dependencies. Question/issues:

- Should the NET Core dependencies be included in my installer? Pointer to guidelines?
- Can I avoid shipping these dependencies if I make NET Core a install a prerequisite (possibly a stupid question)
- I'm new to NET Core, I'm using NET Framwork ... any known issues using within a NET Framework 4.6.1 application?

Thanks

nemec:
Hi, yeah it does install a few 'core' dependencies even on the full framework. It's unintentional and something that needs to be fixed.

In theory you should be able to delete those dependencies from your project manually and it should still work since we aren't using any new APIs in Core. Either way, it definitely works on 4.6.1 without issue (the Nuget is dual-targeted to Core and the Full Framework)

RobSmyth:
Hi nemec,

Thanks, that helps me put it in perspective. For your FYI, for users like me, it would be much better to have separate Core/Framework packages as we do not, in production, reference NuGet web sites. So, in this case I had to download and install all of the Core packages into our local NuGet site. I also have to justify the extra NuGet references in code reviews.

So it was/is quite a burden that I have not through yet. I may need to revert if other reject the code as it adds Core packages.

Thx

Rob

ericnewton76:
@RobSmyth hey rob, I dont think you need to include Core dependencies (better to use the term "netstandard" now) in your installer since I'm assuming you're project targets a desktop framework, ie, netframework 4.6.1+ then you should be okay.  That also means you're not invoking your program  via `dotnet run`.   

Btw, we're not the only package that includes both netstandard/netframework dependencies, however depending on your nuget hosting situation, you may have had to download the netstandard dependencies just to make the package valid on your own nuget server.  Based on the current direction of nuget and netcore, I think you'll find this to be the case for more libraries going forward.

Also, certain netframework versions support particular netstandard versions, check the netstandard support matrix on msdn for more info: https://github.com/dotnet/standard/blob/master/docs/versions.md

Our design is intended to be the lowest common denominator that we could use, which initially was netstandard1.5 which is also supported by netframework 4.6.1+

okvalsralos:
@RobSmyth 
In my humble opinion the NuGet package design is wrong and causes unnecessary confusions.
Properly created NuGet package should look like this:
[...]

Issue #222, opened 2018-01-14
Commandline Nuspec targets prerelease netstandard15 libraries


ericnewton76:
We should upgrade these and potentially separate out net45/net46/net47 to not need to include these (since they are part of desktop framework.  Addresses thoughts brought up with #220 

```
    <dependencies>
      <group targetFramework=".NETStandard1.5">
        <dependency id="System.Collections" version="4.0.11-rc2-24027" />
        <dependency id="System.Console" version="4.0.0-rc2-24027" />
        <dependency id="System.Diagnostics.Debug" version="4.0.11-rc2-24027" />
        <dependency id="System.Globalization" version="4.0.11-rc2-24027" />
        <dependency id="System.IO" version="4.1.0-rc2-24027" />
        <dependency id="System.Linq" version="4.1.0-rc2-24027" />
        <dependency id="System.Linq.Expressions" version="4.0.11-rc2-24027" />
        <dependency id="System.Reflection" version="4.1.0-rc2-24027" />
        <dependency id="System.Reflection.Extensions" version="4.0.1-rc2-24027" />
        <dependency id="System.Reflection.TypeExtensions" version="4.1.0-rc2-24027" />
        <dependency id="System.Resources.ResourceManager" version="4.0.1-rc2-24027" />
        <dependency id="System.Runtime" version="4.1.0-rc2-24027" />
        <dependency id="System.Runtime.Extensions" version="4.1.0-rc2-24027" />
      </group>
    </dependencies>
```

nemec:
We should probably upgrade to .net core 2.0 while we're at it. We're still using project.json which has been dead for a while. Add an empty targetframework for 45, 46,and 47 and hopefully NuGet will install only the needed depencencies.

ericnewton76:
Ok, I'll play with that sometime this week.  

Do we get any benefits from going to NetStandard20?   Seems like NetStandard15 works for now unless you know of some functionality missing

nemec:
Our .net core project format (.xproj, project.json, etc.) are not supported by VS 2017. I know 1.5 came out before the new .csproj project format, so I'm not sure if we can upgrade and keep 1.5 support at the same time. If we can, no problem, but we should definitely upgrade away from project.json so that we get the latest tooling support during development.

PhilPJL:
@ericnewton76 
The benefit for me (and other .NET 4.x) users is that when we reference CommandLineParser we aren't forced to reference a load of .NETStandard packages unnecessarily.

PR #257, opened 2018-03-20
Fix support for .NET 4.0 and .NET 4.5

Athari:
Covers #222 and #227.

I would have fixed .nuspec more throroughly, but I can't even get .NET Standard 1.5 library to build, so this is the only change I can do reliably without breaking anything.

See https://github.com/commandlineparser/commandline/issues/227#issuecomment-373979501 for more details on the issues with .nuspec file and the following comment with a way to fix it.

### DIFF
--- a/CommandLine.nuspec
+++ b/CommandLine.nuspec
@@ -30,6 +30,8 @@
         <dependency id="System.Runtime" version="4.1.0-rc2-24027" />
         <dependency id="System.Runtime.Extensions" version="4.1.0-rc2-24027" />
       </group>
+      <group targetFramework=".NETFramework4.0" />
+      <group targetFramework=".NETFramework4.5" />
     </dependencies>
   </metadata>
   <files>


ericnewton76:
This looks valid.  

I need to figure out how to get Appveyor to do a little more related to these types of changes for smoke testing.  I'm thinking example programs that are targeting Framework v4.0 and Framework v4.5,v4.6,v4.7 and so on and have appveyor/ci build run nuget install or update -r on these example programs to see if theres some install dependency problems.

Did you already create two test apps that are targeting Fx v4.0 and v4.5?  Maybe send a pull request on the examples repo to add those, under targeting_tests subdirectory.  

I mentioned using the examples repo because my first instinct is let them version independently from the source repo (commandline) to help simulate real world scenarios

Athari:
@ericnewton76 I haven't tested anything, I just copied the required lines from another .nuspec.

I can't get the current build scripts to work properly and don't have access to CI artifacts (if AppVeyor publishes .nupkg), so I can't actually test anything.

If you give me carte blanche on the build and project system, I can perform the following changes:

1. replace Paket dependency manager with NuGet;
1. remove dependencies on single files by moving them to repo itself;
1. convert .csproj to new SDK format with multi-targeting and automatic NuGet packaging;
1. remove strong assembly naming and possibly add it as a separate package;
1. update AppVeyor scripts accordingly.

While smoke testing is a good thing, these changes alone would:

1. produce reliable NuGet packages as .nuspec would be generated from project files instead of being written by hand;
1. make the build system more accessible to other developers as no third-party tools would be required;
1. make the package follow the usual recommendations (including not signing);
1. bump minimum VS version requirement to VS2017, where new .csproj format is supported.

I consider the code of the library itself to be in the "doomed" state, as it's written in a hacky F#-in-C# style which nobody except the original author uses and depends on a low-quality base library, and replacing all "maybe" nonsense with normal code, fixing all IEnumerable usages, getting rid of all pointless inverse-of-control and pattern matching imitation throughout API etc. would be a major effort. However, some things can be improved without major changes and I think it's worth the effort as the library is still very popular.

gbritton1:
+1 except for vs 2017 dependency.  My team will be on 2015 for a while. 

Currently though I can't build the solution which is a real drag. Specifically the test project.

Athari:
@gbritton1 VS2017 is a library development dependency. NuGet and app dependencies remain the same. Now that fully featured VS Community is free, I see no point in not switching to new VS version in relatively small open-source projects where it can be done without any hassle.

The rest of changes can't be done without switching to VS2017 anyway. Multi-tartgeting and NuGet package generation on build are features of new SDK based project system of VS2017.

gbritton1:
@Athari well then, let's get on with it!

Lakritzator:
I was really excited about commandlineparser 2.3.0 arriving yesterday!
... until I found out that this change is not yet in there.

What is the current plan for this?

I have the same issue like many others, trying to release a .NET framework 4.6.1 application, but using this package forces many additional unneeded system dependencies which bloat the application.

ericnewton76:
We're close to releasing a NetStandard20 based.  A Microsoft fellow (Viktor Hofer) did some work to bring the library to that level which minimizes the required dependencies from the NetStandard side.

On the NetFramework side, should we push it back to targeting v4.0?  Would that support this need better?

UPDATE: I just realized this PR does add that.  I'm (re-)looking into it now.

Lakritzator:
@ericnewton76 This is about the nuspec only targeting netstandard 1.5 and not net40 / net45.
This result is that the nuget package is used at netstandard 1.5, and when you use 4.6.1 this causes the netstandard dependencies to be added.

As the project seems to be for net40, this doesn't make sense.

This PR should fix the targeting issue, which is a simple change and solves an issue which will help a lot of people.

ericnewton76:
Ok, I gotcha.  I should've included this in v2.3.  I will bump to v2.4 and probably deprecate v2.3.0

jsreynolds:
Any update on this by chance? I had to change the .nuspec on the 2.3 to get it to play nicely with .Net 4.6.1. 

Thanks,

--J

PR #264, opened 2018-03-27
Fix FormatName.  Short name not preferred?  Go with long!  No short name? Has to be long!

gbritton1:
Currently, when unformatting parsed options, if preferShortName is requested but there is not short name (length == 0) then a blank is printed.  This patch will take the long name in that case.


gbritton1:
Back to the drawing board

PR #268, opened 2018-04-02
Show hidden

gbritton1:
Currently, the UnParser returns a command line with hidden options exposed.  I believe that this is wrong.  This pull request changes that behavior:

1. By default, the unparser will keep hidden options hidden
2. I added a new UnParserSetting option, showHidden, to expose hidden options if you really want to

Of course, for any users expecting  hidden options to be exposed by the unparser, this would be considered a breaking change.  If that is unacceptable, it is a small thing to change the option to "hideHidden" to reverse the behavior.


ericnewton76 (approved):
looks good. thanks for the contribution

PR #276, opened 2018-04-19
fixing grammar error in specification error message

DillonAd:
Ran into this error message because I tried to use a single character variable name. The syntax confused me and this is my proposed fix.


ericnewton76:
Thanks!

PR #278, opened 2018-04-23
Fix to Display Width Tests

Oddley:
Explicitly passing MaximumDisplayWidth = 80 to tests that fail unless your Maximum Display Width is 80


ericnewton76 (approved):
This looks good.  I've run into this problem before with the tests and display width being different on my machine, vs appveyor, and another user found out that running inside a default docker container would get different results as well.

LeslieMurphy:
I did some testing - looks like this issue actually got fixed a while back.  It works as of 2.4 and later.  

Fails on 2.3 and earlier with exception System.ArgumentOutOfRangeException: Specified argument was out of the range of valid values. (Parameter 'value')

Issue #277, opened 2018-04-20
Crash when computed help text larger than maximum length


Chadehoc:
Maximum length for displaying help is 120 in `Parser.Default`. If a long help text makes the computed length larger than that (when reporting a parsing error), an `ArgumentOutOfRangeException` is raised.

Details:

In `HelpText.AddOptionsImpl`, `maxLength = GetMaxLength(specifications)` is large enough so that `remainingSpace` below becomes negative, which is passed to `HelpText.AddOption` as `widthOfHelpText`, and used there with an exception.

If I manually set the negative `remainingSpace` to 0 in the debugger, it goes on to an out of memory error. Setting to 1 is ok, but illogical, so there must be another hidden bug that 0 reveals.

Also, this error only occurs with `Parser.Default`. If I use an unmodified `new Parser(s => Console.WriteLine(s.MaximumDisplayWidth))`, this prints 120, the same result is returned (error list), but without crashing. So this behaviour also depends on another setting than the maximum length.

Oddley:
I was also seeing ArgumentOutOfRangeException related to MaximumDisplayWidth. It seems that old versions of mono for non-Windows systems would throw when Console.Width is called, but for newer versions of Mono this is not the case. See #279

moh-hassan:
Fixed in V2.7+

PR #279, opened 2018-04-23
Fixing DisplayWidth for newer Mono

Oddley:
ParserSettings and HelpText assume that Console.WindowWidth throws for non-Windows environments, but this is not true in more recent versions of Mono (returns 0 instead). This change will cause it to default to DefaultMaximumLength in the event that Console.WindowWidth returns 0.

### DIFF
--- a/src/CommandLine/ParserSettings.cs
+++ b/src/CommandLine/ParserSettings.cs
@@ -35,6 +35,10 @@ namespace CommandLine
             try
             {
                 maximumDisplayWidth = Console.WindowWidth;
+                if (maximumDisplayWidth < 1)
+                {
+                    maximumDisplayWidth = DefaultMaximumLength;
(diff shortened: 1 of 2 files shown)


Oddley:
I'm fairly certain this fixes #277. The issue I was seeing and the issue described are extremely similar.

ericnewton76 (approved):
makes sense! thanks

Oddley:
@ericnewton76 Any ETA for when 2.3.0 is getting released? I have a project that would benefit from it.

chsalgado:
Hi, we are consuming this package as it is a great multiplat option, however, we have been affected by this bug when running inside a Linux container, I think this PR would address our issues and would like to know if there is any ETA to ship it, as 2.3.0 came and go without it.

Issue #283, opened 2018-04-27
Can I add validate feature in Set parameter, and throw exception, and show usage?


Wenduan-Mou:
Can I add validate feature in Set parameter, and throw exception, and show usage?
is there exception handling in command line parameter parsing?

[Option(
'f', "filename", 
HelpDesk = "Specifies the input file")]
   public string Name
   {
      get { return mName; }
      set
      {
         Regex regex = new Regex("^[0-9a-z]+.txt$");

         if (regex.Match(value).Success )
            throw new InvalidOptionValueException(
                "The name must be txt file", false);
         mName = value;
      }
   }
 Private string mName;

e673:
No, there is no such exception handling.
Try this code:

```csharp
try
{
    var result = Parser.Default.ParseArguments(...);
}
catch (InvalidOperationException e)
{
    if (e.InnerException is TargetInvocationException targetInvocationException)
    {
        var innerException = targetInvocationException.InnerException;
        Console.Error.WriteLine(innerException.Message);
    }
}
```

P.S. I will try to implement this feature now.

e673:
I've implementing it here:
https://github.com/e673/commandline/tree/issue283
I will make a pull request after adding tests.

ericnewton76:
You can go ahead and create the pull request, we'll simply hold on it as a matter of policy of needing tests anyways.

Thanks for the work!

mmajcica:
I just tried this and it works as expected! Thanks

PR #286, opened 2018-05-05
Implement new functionality suggested in Issue 283

e673:
When a property throws an exception when settings a value, it is treated as parsing error.


ericnewton76:
This pull request addresses Issue #283

Just adding this comment so Github actually links them together properly

ericnewton76 (approved):
Great work! Thanks a bunch.  I'll let it sit for a bit to gain exposure and then merge it in.

Issue #34, opened 2017-11-04
HelpText.AutoBuild ignores word after last dot in AssemblyTitleAttribute


ericnewton76:
<a href="https://github.com/eakoning"><img src="https://avatars0.githubusercontent.com/u/1950705?v=4" align="left" width="96" height="96" hspace="10"></img></a> **Issue by [eakoning](https://github.com/eakoning)**
_Wednesday Jul 15, 2015 at 14:24 GMT_
_Originally opened as https://github.com/gsscoder/commandline/issues/199_

----

I'm using CommandLine version 1.9.71.2
If i specify an `AssemblyTitleAttribute` with a dotted name (and I version my tool 1.10.1.\* through the AssemblyVersionAttribute) in the project's AssemblyInfo.cs file, e.g.

``` csharp
[assembly: AssemblyTitle("my.favorite.tool")]
```

the Copyright and Heading properties on HelpText when using the AutoBuild show
`my.favorite 1.10.1.27342`

Changing the assembly title to 

``` csharp
[assembly: AssemblyTitle("my.great.fantastic.favorite.tool")]
```

changes Copyright and Heading properties to
`my.great.fantastic.favorite 1.10.1.27342`

It always leaves out the word after the last dot.

When I leave out the `AssemblyTitleAttribute` all together, the HelpText falls back to the text specified in the `AssemblyProductAttribute` and that seems to be working fine.

ericnewton76:
<a href="https://github.com/gsscoder"><img src="https://avatars3.githubusercontent.com/u/1194228?v=4" align="left" width="48" height="48" hspace="10"></img></a> **Comment by [gsscoder](https://github.com/gsscoder)**
_Thursday Jul 16, 2015 at 03:25 GMT_

----

@eakoning, can you test it with **2.0**. Changes to public API are easy even if you use it with multiple verb classes.

The effort to maintain two branch is great for us, but if the problem persists in **2.0** I'll be glad to fix it.

Thank you.

ericnewton76:
<a href="https://github.com/scadorel"><img src="https://avatars3.githubusercontent.com/u/2738119?v=4" align="left" width="48" height="48" hspace="10"></img></a> **Comment by [scadorel](https://github.com/scadorel)**
_Friday Oct 16, 2015 at 20:50 GMT_

----

Reproduced using latest beta release (2.0.275.0-beta).
Word after the last dot is removed.

[...]

PR #312, opened 2018-08-02
Fix not showing correct header info, issue #34

tynar:
Added preprocessor for .NET Standard and .NET Framework to fix HeadingInfo issue #34 .

### DIFF
--- a/src/CommandLine/Text/HeadingInfo.cs
+++ b/src/CommandLine/Text/HeadingInfo.cs
@@ -55,10 +55,19 @@ namespace CommandLine.Text
         {
             get
             {
-                var title = ReflectionHelper.GetAttribute<AssemblyTitleAttribute>()
+                string title = ReflectionHelper.GetAssemblyName();
+#if NETSTANDARD1_5
+                title = ReflectionHelper.GetAttribute<AssemblyTitleAttribute>()
                     .MapValueOrDefault(
                         titleAttribute => Path.GetFileNameWithoutExtension(titleAttribute.Title),
-                        ReflectionHelper.GetAssemblyName());
+                        title);
+#else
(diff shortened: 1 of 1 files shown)


nemec:
I'm beginning to think a better solution may be to simply remove `.dll` from the end of the title if it exists. It won't require adding new preprocessors and will only affect those who intentionally end their title with `.dll`, for whatever reason, instead of all of NETSTANDARD.

@tynar I don't see a preprocessor for .NET Framework that calls `Path.GetFileNameWithoutExtension`, it would hit your `#else` and simply not trim. Since that trim call was added back in 2013, before Core even existed, .NET Framework users should get the behavior that trims `.dll` off the end as well as netstandard1_5.

Could you rewrite your change to remove just the string `.dll` from the end of the title, if it exists? If it passes the tests after that, we should be good to go.

tynar:
Wasn't the intention to show the title from AssemblyInfo rather than looking for executed file name? That's why I left else part with title info which is available in .NET Framework.

nemec:
Yes. Note that the existing code calls `GetFileNameWithoutExtension` on the title attribute, NOT the assembly name. I believe it's meant to solve a problem where, by default, the title attribute is given a name of `something.dll` if you did not manually set it yourself.

I think if you take the current published code and then replace

    GetFileNameWithoutExtension(titleAttribute.Title)

 with the following pseudocode (sorry I don't have the time to format it properly)

    title = titleAttribute.Title
    if title endswith ".dll"
        title = title.Substring(0, title.Length - ".dll".Length)

It should solve the original issue while keeping the spirit of the old behavior.

nemec:
That looks good to me @tynar, thank you for your contribution to the project!

Issue #321, opened 2018-09-01
Localization and SentenceBuilder


SpawnProd:
In current version we can't localize a few things.
"Default" word, "Valid values", "pos.", "value pos."...
Everything we can put in SentenceBuilder I suppose.
Also it's not clearly for me, why you use RequireWord as parameter (in AddOptionsImpl function for example).

SpawnProd:
I can do it by myself (just finished), but can I change sdk version for .Net Core build? Current version "1.0.0-preview2-003121" a little bit ancient...
Also, for nupkg in CommandLine.nuspec will be great to change dependency version from "4.0.11-rc2-24027" to "4.3" (without changing .Net Standard). (Issue #297 )

moh-hassan:
Localization is supported in v2.6.0
A localization demo is [available](https://github.com/commandlineparser/commandline/tree/master/demo/ReadText.LocalizedDemo)

Issue #297, opened 2018-06-08
Update system dependencies


Dundret:
You cannot currently use the dotnet publish command with a specific runtime identifier because of the dependency to 4.1.0-rc2-24027 and needs to be updated.

Error messages: 
error NU1605: Detected package downgrade: System.IO from 4.3.0 to 4.1.0-rc2-24027.
error NU1605: Detected package downgrade: System.Resources.ResourceManager from 4.3.0 to 4.0.1-rc2-24027
error NU1605: Detected package downgrade: System.Text.Encoding from 4.3.0 to 4.0.11-rc2-24027.
error NU1605: Detected package downgrade: System.Reflection from 4.3.0 to 4.1.0-rc2-24027.

And so on...

Are these perhaps going to be updated when the update to target .net standard 2.0 occurs?

PR #307, opened 2018-07-20
Move to .NET Standard 2.0

ViktorHofer:
Fixes https://github.com/commandlineparser/commandline/issues/284

I'm from the .NET team and we are currently using this package in of our repos `xunit-performance`. As we are facing some package downgrades with the existing nuget package I upgraded it to target NS2.0.

Changes:
- src assembly `commandline.dll` now targets netstandard2.0 and the xunit test library `commandline.tests` now targets netcoreapp2.0 (not 2.1 as I wasn't sure if the CI already uses the 2.1 SDK)
- Updated dependencies (xunit, FluentAssertions)
- Enable test run in CI. After updating the dependencies 20 tests failed because they are asserting of string which changed by xunit. I recommend removing such flaky tests. I disabled them for now: https://github.com/commandlineparser/commandline/pull/307/commits/f151ae1c35ab094387cd3d2409bbcd7135df441a
- Replace Paker with the new in-built PackageReference and copy the 4 used files manually into the project to avoid adding new nuget dependencies
- Replace nuspecs with the new in-built csproj attributes
- Converge Test projects
- Update build scripts

Requirements:
- Visual Studio 2017
- .NET Core SDK >= v2.0

Please let me know if you have any questions.

PR #322, opened 2018-09-05
Localization and nuspec

SpawnProd:
Changes in CommandLine.nuspec - dependencies version was changed to 4.3
Additional localization for Default and pos. words, Valid values and value pos phrases.
Also required word as function parameter was removed from some private functions and it was replaced with invoking SentenceBuilder instance functions.
Issue #321 
Issue #297

### DIFF
--- a/CommandLine.nuspec
+++ b/CommandLine.nuspec
@@ -16,19 +16,19 @@
     <tags>command line commandline argument option parser parsing library syntax shell</tags>
     <dependencies>
       <group targetFramework=".NETStandard1.5">
-        <dependency id="System.Collections" version="4.0.11-rc2-24027" />
-        <dependency id="System.Console" version="4.0.0-rc2-24027" />
-        <dependency id="System.Diagnostics.Debug" version="4.0.11-rc2-24027" />
-        <dependency id="System.Globalization" version="4.0.11-rc2-24027" />
-        <dependency id="System.IO" version="4.1.0-rc2-24027" />
-        <dependency id="System.Linq" version="4.1.0-rc2-24027" />
-        <dependency id="System.Linq.Expressions" version="4.0.11-rc2-24027" />
-        <dependency id="System.Reflection" version="4.1.0-rc2-24027" />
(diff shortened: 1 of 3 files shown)


CommandLine.nuspec
```diff
@@ -16,19 +16,19 @@
...
+        <dependency id="System.IO" version="4.3.0" />
+        <dependency id="System.Linq" version="4.3.0" />
+        <dependency id="System.Linq.Expressions" version="4.3.0" />
+        <dependency id="System.Reflection" version="4.3.0" />
+        <dependency id="System.Reflection.Extensions" version="4.3.0" />
+        <dependency id="System.Reflection.TypeExtensions" version="4.3.0" />
+        <dependency id="System.Resources.ResourceManager" version="4.3.0" />
+        <dependency id="System.Runtime" version="4.3.0" />
+        <dependency id="System.Runtime.Extensions" version="4.3.0" />
       </group>
```
ericnewton76, line 32: dependency changes will probably conflict with upcoming netcore standard 2.0 adjustments (#307)

src/CommandLine/Text/HelpText.cs
```diff
@@ -860,8 +855,8 @@ private string AddValueName(int maxLength, ValueSpecification specification)
             return new StringBuilder(maxLength)
                 .BimapIf(
                     specification.MetaName.Length > 0,
-                    it => it.AppendFormat("{0} (pos. {1})", specification.MetaName, specification.Index),
-                    it => it.AppendFormat("value pos. {0}", specification.Index))
+                    it => it.AppendFormat("{0} ({1} {2})", specification.MetaName, SentenceBuilder.PositionWord(), specification.Index),
+                    it => it.AppendFormat("{0} {1}", SentenceBuilder.ValuePositionPhrase(), specification.Index))
```
ericnewton76, line 859: I think these localization changes dont actually fit the need.

Some languages might need the number in a different spot.  French sometimes likes '1 de juen' instead of 'june 1' for example (thats a date but you should understand my point).  

The strings might need to include numeric placeholders too... I feel like localization for now needs to be reconsidered and rearchitected in this lib.

PR #325, opened 2018-09-07
Add a link to the nuget package in README

dsschnau:
Just to make life a bit easier


nemec:
Thank you for your contribution @dsschnau! The NuGet badge at the top of the readme also contains a link to the package, but clearly it is not an obvious place to find it :)

PR #343, opened 2018-10-04
[WIP][Tests] Fix locally failing tests 

Yiabiten:
Do not assume Console width in contributors machines
(in my situation: Win10 with console width=120)

I fixed all failing tests: basically all tests with some option help-text's length higher than 80 (that are expected to be wrapped) but I think this should be done for all tests.

### DIFF
--- a/tests/CommandLine.Tests/Unit/ParserTests.cs
+++ b/tests/CommandLine.Tests/Unit/ParserTests.cs
@@ -488,6 +488,7 @@ namespace CommandLine.Tests.Unit
         [Fact]
         public void Properly_formatted_help_screen_is_displayed_when_usage_is_defined_in_verb_scenario()
         {
+            Console.WindowWidth = 80;
             // Fixture setup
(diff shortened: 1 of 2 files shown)


ericnewton76:
This issue has caused problems for several.

The other prblem is it works on your local but the build server does something different and causes the same tests to fail.

Build server (appveyor) is using redirected output, so calls to GetWindowInfo throw an IOException when the sizing is being determined.  The call stack doesnt help much, only telling the test that failed and not the specific method within the library thats initiating a call to Console.get_WindowHeight : https://ci.appveyor.com/project/commandlineparser/commandline/build/2.2.69-lipauiuy/job/xu37x7acde81g12p#L466

Another fellow reported that Docker seems to do similar as Appveyor, where the console output is probably redirected and would cause an IOException to be thrown.

I'm just not sure what's generating that call to Console.get_WindowHeight.  Searching for any "WindowHeight" strings in the project bring up literally nothing.

Issue #330, opened 2018-09-18
BaseAttribute HelpText localization


tkouba:
Add support for HelpText localization. I want localize whole help (texts, examples), but HelpText can not be localized. 

One of the possible solutions is solution like System.ComponentModel.DataAnnotations.DisplayAttribute with ResourceType property.

Insire:
I'd be interested in a solution for this aswell.

PR #355, opened 2018-10-31
Support for HelpText localization with ResourceType property. Closing issue #330

tkouba:


tkouba:
NetStandard 1.5. build failed

PR #356, opened 2018-10-31
Support for HelpText localization with ResourceType property.

tkouba:
Closing issue #330


ericnewton76:
Looks good, but I have one question... (wihtout looking myself) isn't there already a Localizable attribute in System.ComponentModel that can be used?

Or is the reason for the "new" attribute here in the library, because of Netstandard/netcore/netfx differences?

tkouba:
This localize HelpText property of Option/Value attribute throw resources. This is not same as LocalizableAttribute from System.ComponentModel. Usage is inspired from DisplayAttribute (https://docs.microsoft.com/en-us/dotnet/api/system.componentmodel.dataannotations.displayattribute). Infrastructure helper LocalizableAttributeProperty is used only once for HelpText property.

Usage is:
[Option("lastname", HelpText = "LastName", ResourceType = typeof(EmployeeResx))]

and create EmployeeResx resource, so HelpText will be shown localized.

gkelling:
Is this something that is still being considered?

moh-hassan:
Localization is required  as posted in the issues #321, #361.
Localization here  is related to the application side and  can be created independent of the library localization. It can cover the Properties: HelpText and MetaName.

Localizing  the library can be done by localizing all the constant strings used in the library like "Required" , "Default", .. and so on including the error messages.
This localization can  start by creating the default English resource file, e.g `strings.resx` then contributors can build other language resources e.g strings.fr.resx , string.ru.resx and so on based on the English resource file.

What is your suggestion based on the above text?

gkelling:
I must be missing something.  I was looking at this PR conversation because I was unable to use a resx file with English strings to populate the helptext field in the options attribute, because it is an attribute and resx file values are not constants.

This post seemed to solve that problem, but I'm happy to do it another way if you could help me connect the dots. I can provide a code example of that would help.

moh-hassan:
@gkelling 
Welcome and I can help  in building  the localization.
I'm waiting your example. :)

gkelling:
Using the example above with a few modifications, we have this.  
`[Option("lastname", HelpText = "Enter your last name.")]`

I want to set "Enter your last name." using a resx file.
If I create a resx file named HelpTextStrings.resx with TestValue = Enter your last name.
When I try to do this I get and error stating I must use a constant.
`[Option("lastname", HelpText = HelpTextStrings.TestValue)]`

Any help would be appreciated.

moh-hassan:
The error stating you  must use a constant is expected because it's used in Attribute properties. 

Steps to localize Console application:

- Build your resource files: strings.resx , strings.fr.resx, ...
- Using T4 template,  generate a class from the resource file with const strings, like:

        //*** Auto generated class from file: C:\Users\admin\source\repos\ResDemo\ResDemo\Resource1.resx
        public class Strings
        {
          public const string Msg1 = "Prints all messages to standard output." ;
          public const string Msg2 = "Read from stdin" ;
          public const string TestValue = "Enter your last name" ;
        }

- define HelpText for Option/Value  using these constants:

     class Options
        {
           
            [Option("lastname", HelpText =Strings.TestValue)]
            public string LastName { get; set; }           
      
        }
 
I find PR of  @tkouba define HelpText using the resource file: ResourceType= typeof(EmployeeResx) 
Which may reduce the above steps.

        [Option("lastname", HelpText = "LastName", ResourceType = typeof(EmployeeResx))]



@tkouba       
Can you convert your PR to vs2017 and rebase to develop branch.
It's nice if you can provide a demo console application that use the  ResourceType

gkelling:
@moh-hassan 
I updated the code to use a text template that generates a class of constant strings using the resx file and I use those to set the help text.  That all works.

My concern is that the class is generated when the project is compiled and thus the strings are set to the language that the machine compiling is using.  There is no localization determined at runtime, where as the solution presented by @tkouba would utilize the resx framework for this.

Or were you suggesting that a runtime text template be used?  I was under the impression those were only used for text manipulation, not code generation.

moh-hassan:
The  provided solution (work around) localize the Application at compile time due to the limitation of Attribute'Properties of using constant expression.
I agree with you that the solution presented by @tkouba may utilize the resx framework at runtime.

I start an integration test with this PR  and there are some fail of test cases and I try managing to fix it with the author of other feature.

moh-hassan:
@tkouba
please, Can you convert your PR to vs2017 and rebase to develop branch.
There is fail of 9 test cases as described [here](https://github.com/commandlineparser/commandline/pull/458#issuecomment-508957707).

The last version in develop is 2.5.0 support netstandard2,net461,net45 and net40 and working in vs2017  (Paket not supported).

moh-hassan:
- Rebased with develop
- Resolve confliction
- Fix Test fail by modification:
      Set Helptext property default to string.Empty
-Modify #if NETSTANDARD1_5
- Merge to develop
- Pass CI Appveyor

moh-hassan:
@gkelling
You can try the localization feature using dev package at [AppVeyor](https://ci.appveyor.com/api/buildjobs/a9n4fjbp267itfhs/artifacts/src%2FCommandLine%2Fbin%2FRelease%2FCommandLineParser.2.5.37.nupkg)
The feature will be available in the next release 2.5.6

gkelling:
@moh-hassan 
I updated my repo with the dev package and tested.  Everything is working. Thank you.

Issue #389, opened 2019-01-15
Version 2.4.x  System.ArgumentNullException parsing options in unit-test


mbedded:
Hello,

i tried all 2.4.x versions (2.4.0, 2.4.1, 2.4.2, 2.4.3) and all have the same exception. In my unit-test i call my program only with "--help" after that, i receive the following stacktrace:

```
// Stacktrace
System.ArgumentNullException
Value cannot be null.
Parameter name: element
   at System.Attribute.GetCustomAttributes(Assembly element, Type attributeType, Boolean inherit)
   at System.Reflection.CustomAttributeExtensions.GetCustomAttributes[T](Assembly element)
   at CommandLine.Infrastructure.ReflectionHelper.GetAttribute[TAttribute]()
   at CommandLine.Text.HelpText.AutoBuild[T](ParserResult`1 parserResult, Func`2 onError, Func`2 onExample, Boolean verbsIndex, Int32 maxDisplayWidth)
   at CommandLine.Text.HelpText.AutoBuild[T](ParserResult`1 parserResult, Int32 maxDisplayWidth)
   at CommandLine.Parser.<>c__DisplayClass17_0`1.<DisplayHelp>b__1(IEnumerable`1 _, TextWriter writer)
   at CSharpx.MaybeExtensions.Do[T1,T2](Maybe`1 maybe, Action`2 action)
   at CommandLine.ParserResultExtensions.WithNotParsed[T](ParserResult`1 result, Action`1 action)
   at CommandLine.Parser.DisplayHelp[T](ParserResult`1 parserResult, TextWriter helpWriter, Int32 maxDisplayWidth)
   at Program.<Main>d__5.MoveNext() in C:\..\..\Program.cs:line 33
```

My code snippets are the following ones:
```csharp
// Test method (xUnit) which fails
[Fact]
public async Task CallMain_GiveHelpArgument_ExpectSuccess() {
    var result = await Program.Main(new[] { "--help" });

    Assert.Equal(ERROR_SUCCESS, result);
}
```

```csharp
// main program
internal class Program {

  private const int ERROR_SUCCESS = 0;

  internal static async Task<int> Main(string[] args) {
      bool hasError = false;
      bool helpOrVersionRequested = false;

      ParserResult<Options> parsedOptions = Parser.Default.ParseArguments<Options>(args).WithNotParsed(errors => {
[...]

PR #392, opened 2019-01-18
[no-merge] Add unit tests for Issue389

ericnewton76:
This tries to address #389 

Note, this PR may never get merged.

### DIFF
--- a/tests/CommandLine.Tests/Unit/Issue389_Tests.cs
+++ b/tests/CommandLine.Tests/Unit/Issue389_Tests.cs
@@ -1,0 +1,78 @@
+﻿using CommandLine.Text;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using Xunit;
+
+namespace CommandLine.Tests.Unit
+{
+	
+	public class Issue389_Tests
+	{
+
+		private const int ERROR_SUCCESS = 0;
+
+		// Test method (xUnit) which fails
+		[Fact]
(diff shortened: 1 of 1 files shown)


moh-hassan:
The resolve this issue , the test class should reference the next class:

		 public class BaseTest
			{
				public BaseTest()  
				{  
					//in case multi target project,use #if because no appdomain in netcore
		#if NETFRAMEWORK
				  
					var assembly = Assembly.GetCallingAssembly();  

					var manager = new AppDomainManager();  
					var entryAssemblyfield = manager.GetType().GetField("m_entryAssembly", BindingFlags.Instance | BindingFlags.NonPublic);  
					entryAssemblyfield?.SetValue(manager, assembly); 

					var domain = AppDomain.CurrentDomain;  
					var domainManagerField = domain.GetType().GetField("_domainManager", BindingFlags.Instance | BindingFlags.NonPublic);  
					domainManagerField?.SetValue(domain, manager);  
					
		#endif
				} 
			}

Then test class inherit BaseTest:

          public class Issue389_Tests: BaseTest
	   {
          //test stuff
          }

The solution is tested  for the same issue#389 [here](https://github.com/commandlineparser/commandline/issues/389#issuecomment-458406805)

moh-hassan:
The test case is involved and Pass in PR [389](https://github.com/commandlineparser/commandline/pull/387) in both net461 and netcoreapp2.0  after resolving the null EntryAssemply problems

moh-hassan:
@ericnewton76 
It's added and merged to PR #430

PR #408, opened 2019-02-05
Support of Xamarin.Mac / mono

niveuseverto:
When mono compiled without MONO_FEATURE_CONSOLE define, some methods of System.Console are throw System.PlatformNotSupportedException.

To be exact: [Console.WindowWidth](https://github.com/mono/mono/tree/afd4971208135b8e1f1b11755f531647a4195b9f/mcs/class/corlib/System/Console.cs#L895-L899) in this case.


ajtruckle:
I propose that this one been closed because Xamarin is now superseded by MAUI.

Issue #417, opened 2019-03-12
Why does ParserSettings implement IDisposable?


unageek:
Why does `ParserSettings` implement `IDisposable` while it disposes of nothing?

Is that because it [used to dispose of `HelpWriter`](https://github.com/commandlineparser/commandline/commit/c75a69abadd02daf984fb63e4256e3e4714d8345#diff-00a389261204bf2e245786d84442cfacR181) and you are not able to get rid of the interface without breaking the API?

ericnewton76:
Probably.  Personally I dont know the project that in-depth and that's a reasonable guess.  It could easily be up for elimination in a vNext major release.

PR #423, opened 2019-03-28
Write built-in help/version text to Console.Out when using the default parser.

### LINKED ISSUE #399: Built in Help and Version output goes to stderr not stdout.
The built in --help and --version outputs are printed to stderr instead of stdout, you have to redirect them to use the commands properly.

`test --version` returns the version number, but on stderr not stdout.

e.g. 
`test --version 2>nul` returns no output.
`test --version 2>&1` prints the version to stdout as expected.

twaalewijn:
Fixes #399.

Instead of adding another TextWriter property to ParserSettings I've opted to move the new behavior into a separate class.

This ensures that:
- Disabling all help text via ```settings.HelpWriter = null``` still works.
- Developers assigning their own TextWriter are notified that the help text output can be split into user requested help and parser error triggered help because the HelpWriter class cannot be assigned with a TextWriter so updating the library will break their builds at the assignment code.
- Because their build breaks after updating, developers won't miss help text output they may have expected to be written to their own TextWriter just because they didn't notice a new TextWriter property in the ParserSettings.
- Developers can still get the old behavior where all help text is written to the same TextWriter by changing their code from ```settings.HelpWriter = textWriter``` to ```settings.HelpWriter = new HelpWriter(textWriter)```.

Should this fix be considered adequate then the [wiki page](https://github.com/commandlineparser/commandline/wiki/Generating-Help-and-Usage-information#customizing-help-output) for customizing help output might also need an update.

### DIFF
--- a/src/CommandLine/Parser.cs
+++ b/src/CommandLine/Parser.cs
@@ -2,7 +2,6 @@
 
 using System;
 using System.Collections.Generic;
-using System.IO;
 using System.Linq;
 using CommandLine.Core;
 using CommandLine.Text;
@@ -19,7 +18,7 @@ namespace CommandLine
         private bool disposed;
         private readonly ParserSettings settings;
         private static readonly Lazy<Parser> DefaultParser = new Lazy<Parser>(
-            () => new Parser(new ParserSettings { HelpWriter = Console.Error }));
+            () => new Parser(new ParserSettings { HelpWriter = HelpWriter.Default }));
 
         /// <summary>
         /// Initializes a new instance of the <see cref="CommandLine.Parser"/> class.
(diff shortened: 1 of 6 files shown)


twaalewijn:
This might be a little off-topic for the issue itself but I noticed that the IDisposable implementations on Parser and ParserSettings don't actually dispose anything.

There is a test in ParserSettingsTests called Disposal_does_not_dispose_HelpWriter which checks if calling Dispose on the settings instance doesn't dispose of the (now moved) TextWriter.

I agree that it shouldn't be the ParserSettings instance's job to dispose of the TextWriters but isn't implementing IDisposable a bit misleading then considering nothing ever gets disposed?

Edit: Nevermind, I didn't see issue #417.
But if such a thing was kept to not break existing code, the fix in this pull request might not be acceptable either.

PR #452, opened 2019-06-01
Fix #379, Enum with 0 value not shown on Example

### LINKED ISSUE #379: Enum with 0 value not shown on Example
public class MyOption 
  {
        [Option('t', "type", Required = true, HelpText = "My entity")]
        public EntityType MyEntityType { get; set; }
public enum MyEntityType
    {
        X0,  // Example for X0 is empty. If change this line to X0 =1, everything is ok 
        X1,
        X2
    }

        [Usage(ApplicationAlias = "...")]
        public static IEnumerable<Example> Examples
        {
            get
            {
                foreach (EntityType entityType in Enum.GetValues(typeof(EntityType)))
                {
                    yield return new Example(entityType.ToString(), new MyOption { MyEntityType = entityType});
                }
            }
        }
    }

njqdev:
See #379

The bug seems to be caused by
https://github.com/commandlineparser/commandline/blob/4d69a3898b84c821161b2db948075cd9e162832c/src/CommandLine/UnParserExtensions.cs#L251

returned to
https://github.com/commandlineparser/commandline/blob/4d69a3898b84c821161b2db948075cd9e162832c/src/CommandLine/UnParserExtensions.cs#L124

I ran all tests after the change and none of them failed.

### DIFF
--- a/src/CommandLine/UnParserExtensions.cs
+++ b/src/CommandLine/UnParserExtensions.cs
@@ -121,7 +121,7 @@ namespace CommandLine
                     type.GetSpecifications(
                         pi => new { Specification = Specification.FromProperty(pi),
                             Value = pi.GetValue(options, null).NormalizeValue(), PropertyValue = pi.GetValue(options, null) })
-                where !info.PropertyValue.IsEmpty()
+                where !info.PropertyValue.IsEmpty(info.Specification)
                 select info)
                     .Memorize();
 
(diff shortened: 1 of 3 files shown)


moh-hassan:
+1
That fix the bug

njqdev:
My change caused another issue so you might want to close this.

see https://github.com/commandlineparser/commandline/issues/379#issuecomment-498055220

moh-hassan:
Instead of closing the Pr, you can convert the PR to **WIP (Work In Progress)** by modifying the title and add WIP as a prefix, and take your time to resolve the side effect.

I think these  methods are  needed to be added:

      object  GetDefaultValue(this OptionSpecification option )
      {
        //get the default value from the specification of option
      }

    bool  HasDefaultValue(this OptionSpecification option )
      {
        //if  the option has a default value from specification return  true otherwise return false
      }

and  `IsEmpty()`  can be  based on the computation of these methods.

I see `IsEmpty()` is calculated from the default value of the  BaseType and it should also calculated based on its spec Default.

[L252  ](https://github.com/commandlineparser/commandline/blob/62467e4ef48cb6c40b23d3a39d7415647b3e0e09/src/CommandLine/UnParserExtensions.cs#L252)
        
            if (value is ValueType && value.Equals(value.GetType().GetDefaultValue())) return true;

njqdev:
Sure, I'll give it a try this week or the next

njqdev:
The best I could do was to change it to: 
`if (!spec.Required && value is ValueType && value.Equals(value.GetType().GetDefaultValue())) return true;`
This solves the issue by always including required value types, and skipping non-required ones which have a default value. 


At first I tried replacing the line with this:
```C#
if (spec.HasDefaultValue() && value.Equals(spec.DefaultValue.FromJust())) return true;
```
but it broke some UnParserExtensionsTests.

For example:

UnParsing_instance_returns_command_line
```
Message: Expected string to be equivalent to 
"-i 1 2 3" with a length of 8, but 
"-i 1 2 3 0" has a length of 10.
```
The BaseType default was included even though it wasn't supposed to be.

PR #460, opened 2019-06-16
updated license url

talanc:
Updating the url to fix NuGet "View License".


moh-hassan:
Thanks @talanc for your work.
It's updated in v2.7.82 using file license.

Issue #455, opened 2019-06-06
Help Text Displays 'help' Option/Verb if AutoHelp is 'False'


x789:
### Description
If I configure the parser to not support the implicit option/verb 'help', the help text still contains the option/verb.
(This is possibly related to #414)

### Expected Behavior
I expect that the option/verb is not displayed.

### Environment
- .NET Core 2.2 Application
- CommandLineParser Nuget Package 2.5.0

### How to Reproduce
1. Configure the parser:
    ```
    var parser = new Parser(config =>
      {
        config.AutoHelp = false;
        config.HelpWriter = Console.Out;
      });
    var parserResult = parser.ParseArguments ...
    ```
2. Execute your program with `--help`
    > ERROR(S):
    >  Verb '--help' is not recognized.
    >
    >  --help       Display this help screen.
    >
    >  --version    Display version information.

x789:
I investigated this issue and found a reason for this behavior. The parser-settings are not propagated to the method (`HelpText.AutoBuild`) that actually builds the help text.
Therefore, I think that #414 and #455 have the same origin.

I should be able to provide a patch in the next days...

x789:
Closed [(Reason)](https://github.com/commandlineparser/commandline/pull/461#issuecomment-506968576)

moh-hassan:
@x789 

The `AutoHelp /AutoVersion` properties of `Parser` for disabling AutoHelp /AutoVersion generation.
The `AutoHelp /AutoVersion` properties of `HelpText` for hiding/showing  --help/--version 
They have the same names but have different purpose.

So to hide --help/--version option , these properties should be false in both Parser and HelpText

var parser = new CommandLine.Parser(with=>
					{ 		
							
						with.AutoHelp=false; //disable AutoHelp Generation
						with.AutoVersion=false; //disable AutoVersion Generation						
					}); 

                              //use HelText to generate help without --help/--version options

                               var helpText = HelpText.AutoBuild(Result, h =>
							{					
								//this is working ONLY when parser  AutoHelp /AutoVersion =false
                // it may better if renamed ShowHelp/ShowVersion
								h.AutoHelp=false; //hide --help option
								h.AutoVersion=false;		//hide --version option									
								return h;
							}, 
							e =>e);

 
  It may be better if these properties are renamed in HelpText to be: ShowHelp/ShowVersion 

**How To show help/version** screen

[...]

Issue #357, opened 2018-10-31
Support Repeated Arguments using standard syntax instead of the non-standard and confusing syntax.


duaneking:
Using the newest version, Despite the fact I declared the option as a list in my options class, I'm not able to add multiple options when I use the standard linux command argument patterns, but I am if I use a pattern that breaks compatibility with some accessibility software used by the disabled.

For example the following creates a repeatedoption error:

```
    public class CommandLineOptions
    {
        [Option('c', "contains", Required = false, Min = 1, Max = 255)]
        public IList<string> Contains { get; set; }
    }
```
And the command line:

```
derp.exe  -c "foo" -c "bar"
```
..Results in an error.

But if I look at the code here on github (because this isn't documented anywhere I can find, at all, despite being a critical use case) I can get it to work if I do:

```
derp.exe  -c "foo" "bar"
```

Respectfully, my first thought on seeing this was "This is a critical defect that explicitly neuters the potential of this library".

My personal usage goal is in the style of core linux tools like grep, awk, sed, etc.  So in my mind this should work exactly like the standard linux flow, where either the last argument wins in the case of only needing one on the command,  or in the case of expecting a possible list due to duplicate use of an option, a list should be built up in the order the command line options are used.

Now we can clearly see that this does work, its just using a very inconsistent syntax that doesn't take into account the needs of screenreaders (So this library technically discriminates against the disabled.  Did you know this?) or consistent linux users - look at the standard grep tool, you can use the same argument multiple times and it just works - but only after looking at the source code could I hunt this down because I could not find any data on repeated options or this confusing and inconsistent syntax anywhere in the docs.

Looking at the code, the tests explicitly do not support the industry standard syntax.   This is even more frustrating, because its effectively giving the finger to people who want to support accessibility tools that DEPEND on this classic, standard, "last one wins" argument parsing.

This lack of support makes the library unusable for most of the projects I had planned to use it in, and while that hurts me personally, I'm actually more worried about all the people who cant use software that uses this library for argument parsing because of the nonstandard syntax due to a disability that chains them to accessability tools that by default wont understand this non-standard syntax.

So respectfully, I feel like you are shooting yourself in the foot not supporting this, and I would formally like to ask that you support the standard syntax so that more people can actually use this.

duaneking:
Updated the title as it was from a prior iteration of what I wrote from when I was testing this to make sure I had all the facts.

nemec:
[...]

Issue #294, opened 2018-05-25
Hide "version" and "help" commands from display


pestebansiam:
Is there any way to hide or remove "version" and "help" commands? 
I just want my console to display my verbs and i haven't seen how to do it.

Thanks.

luissagasta:
It would be very interesting.
That verbs didn't appear in version 1.x

luissagasta:
There is a pending pull request to fix this issue #256

Issue #39, opened 2017-11-04
Displaying help text with fluent syntax?


ericnewton76:
<a href="https://github.com/jawn"><img src="https://avatars3.githubusercontent.com/u/1705112?v=4" align="left" width="96" height="96" hspace="10"></img></a> **Issue by [jawn](https://github.com/jawn)**
_Wednesday Aug 12, 2015 at 08:58 GMT_
_Originally opened as https://github.com/gsscoder/commandline/issues/223_

----

Using the latest nuget of the version 2.0 pre-release, I can display help text like this:

```
var results = Parser.Default.ParseArguments<Options>(args)
  .WithParsed(opts => /* handle options here */)

if (results.Tag == ParserResultType.NotParsed)
{
  var helpText = HelpText.AutoBuild(results);
  Console.WriteLine(helpText);
}
```

Is there a way to do this within the fluent syntax?

Something like
    var results = Parser.Default.ParseArguments<Options>(args)
      .WithNotParsed(opts => Console.WriteLine(HelpText.AutoBuild()));
      .WithParsed(opts => /\* handle options here */)

Since AutoBuild() requires results, this can't be called within the fluent syntax. That's already logged as #88, however I would like to know whether a different solution is available.

ericnewton76:
<a href="https://github.com/gsscoder"><img src="https://avatars3.githubusercontent.com/u/1194228?v=4" align="left" width="48" height="48" hspace="10"></img></a> **Comment by [gsscoder](https://github.com/gsscoder)**
_Wednesday Aug 12, 2015 at 13:23 GMT_

----

Hi @jawn, happy you choose this library.

If you invoke the parser using default singleton like in posted sample, you don't need any call to `AutoBuild()`. This is because this instance is pre-configured with a `ParserSettings.HelpWriter` set to `Console.Error` and when this property has a value the help screen is automatically generated and displayed.

In this case just stay with the `WithParsed` and job is done.

_If you want handle help text manually, there's various way to do it from using `AutoBuild()` to mixing various intermediate step to get whatever customization you want._

Any way a **possible** fluent version, could be an extension to `ParserResult` like `result.AutoBuild()`...

If this not helped, please let me know.

PS: If possible use `ParserResult` extension methods instead of manually checking `Tag` discriminator and cast the instance.

moh-hassan:
Closed for age.

PR #461, opened 2019-06-17
Propagate parser-settings to HelpText (fix #414 and #455)

### LINKED ISSUE #414: Incorrect help text if AutoVersion = false
If you configure the Parser with ParserSettings.AutoVersion=false, the help text still displays the line:
 --version Display version information.

x789:
This fixes issues #414 and #455 (value of `AutoHelp` and `AutoVersion` are ignored by `HelpText`)
This fix does not change the public interface. It is achieved by overloading `HelpText.AutoBuild`. If a `ParameterSettings` object is passed, the values it contains are used. If no `ParameterSettings` are passed to `AutoBuild`, help and version are always displayed.

### DIFF
--- a/src/CommandLine/Parser.cs
+++ b/src/CommandLine/Parser.cs
@@ -194,16 +194,15 @@ namespace CommandLine
         {
             return DisplayHelp(
                 parserResult,
-                settings.HelpWriter,
-                settings.MaximumDisplayWidth);
+                settings);
         }
 
-        private static ParserResult<T> DisplayHelp<T>(ParserResult<T> parserResult, TextWriter helpWriter, int maxDisplayWidth)
+        private static ParserResult<T> DisplayHelp<T>(ParserResult<T> parserResult, ParserSettings settings)
         {
             parserResult.WithNotParsed(
                 errors =>
-                    Maybe.Merge(errors.ToMaybe(), helpWriter.ToMaybe())
-                        .Do((_, writer) => writer.Write(HelpText.AutoBuild(parserResult, maxDisplayWidth)))
+                    Maybe.Merge(errors.ToMaybe(), settings.HelpWriter.ToMaybe())
+                        .Do((_, writer) => writer.Write(HelpText.AutoBuild(parserResult, settings)))
                 );
 
             return parserResult;
--- a/src/CommandLine/Text/HelpText.cs
+++ b/src/CommandLine/Text/HelpText.cs
@@ -236,13 +236,38 @@ namespace CommandLine.Text
             bool verbsIndex = false,
             int maxDisplayWidth = DefaultMaximumLength)
         {
+            var settings = new ParserSettings() { MaximumDisplayWidth = maxDisplayWidth };
+            return HelpText.AutoBuild<T>(parserResult, onError, onExample, settings, verbsIndex);
+        }
+
+        /// <summary>
+        /// Creates a new instance of the <see cref="CommandLine.Text.HelpText"/> class using common defaults.
+        /// </summary>
+        /// <returns>
+        /// An instance of <see cref="CommandLine.Text.HelpText"/> class.
+        /// </returns>
+        /// <param name='parserResult'>The <see cref="CommandLine.ParserResult{T}"/> containing the instance that collected command line arguments parsed with <see cref="CommandLine.Parser"/> class.</param>
+        /// <param name='onError'>A delegate used to customize the text block of reporting parsing errors text block.</param>
+        /// <param name='onExample'>A delegate used to customize <see cref="CommandLine.Text.Example"/> model used to render text block of usage examples.</param>
+        /// <param name="verbsIndex">If true the output style is consistent with verb commands (no dashes), otherwise it outputs options.</param>
+        /// <param name="settings">The parser settings</param>
+        /// <remarks>The parameter <paramref name="verbsIndex"/> is not ontly a metter of formatting, it controls whether to handle verbs or options.</remarks>
+            public static HelpText AutoBuild<T>(
+            ParserResult<T> parserResult,
+            Func<HelpText, HelpText> onError,
+            Func<Example, Example> onExample,
+            ParserSettings settings,
+            bool verbsIndex = false)
+        {
             var auto = new HelpText
             {
                 Heading = HeadingInfo.Empty,
                 Copyright = CopyrightInfo.Empty,
                 AdditionalNewLineAfterOption = true,
                 AddDashesToOption = !verbsIndex,
-                MaximumDisplayWidth = maxDisplayWidth
+                MaximumDisplayWidth = settings.MaximumDisplayWidth,
+                AutoHelp = settings.AutoHelp,
+                AutoVersion = settings.AutoVersion
             };
 
             try
@@ -309,22 +334,39 @@ namespace CommandLine.Text
         /// of <see cref="CommandLine.ParserSettings"/>.</remarks>
         public static HelpText AutoBuild<T>(ParserResult<T> parserResult, int maxDisplayWidth = DefaultMaximumLength)
         {
+            var settings = new ParserSettings() { MaximumDisplayWidth = maxDisplayWidth };
+            return HelpText.AutoBuild<T>(parserResult, settings);
+        }
+
+        /// <summary>
+        /// Creates a new instance of the <see cref="CommandLine.Text.HelpText"/> class,
+        /// automatically handling verbs or options scenario.
+        /// </summary>
+        /// <param name='parserResult'>The <see cref="CommandLine.ParserResult{T}"/> containing the instance that collected command line arguments parsed with <see cref="CommandLine.Parser"/> class.</param>
+        /// <param name="settings">The parser settings.</param>
+        /// <returns>
+        /// An instance of <see cref="CommandLine.Text.HelpText"/> class.
+        /// </returns>
+        /// <remarks>This feature is meant to be invoked automatically by the parser, setting the HelpWriter property
+        /// of <see cref="CommandLine.ParserSettings"/>.</remarks>
+        public static HelpText AutoBuild<T>(ParserResult<T> parserResult, ParserSettings settings)
+        {
             if (parserResult.Tag != ParserResultType.NotParsed)
                 throw new ArgumentException("Excepting NotParsed<T> type.", "parserResult");
 
             var errors = ((NotParsed<T>)parserResult).Errors;
 
             if (errors.Any(e => e.Tag == ErrorType.VersionRequestedError))
-                return new HelpText(HeadingInfo.Default){MaximumDisplayWidth = maxDisplayWidth }.AddPreOptionsLine(Environment.NewLine);
+                return new HelpText(HeadingInfo.Default){MaximumDisplayWidth = settings.MaximumDisplayWidth, AutoHelp = settings.AutoHelp, AutoVersion = settings.AutoVersion }.AddPreOptionsLine(Environment.NewLine);
 
             if (!errors.Any(e => e.Tag == ErrorType.HelpVerbRequestedError))
-                return AutoBuild(parserResult, current => DefaultParsingErrorsHandler(parserResult, current), e => e, maxDisplayWidth: maxDisplayWidth);
+                return AutoBuild(parserResult, current => DefaultParsingErrorsHandler(parserResult, current), e => e, settings);
 
             var err = errors.OfType<HelpVerbRequestedError>().Single();
             var pr = new NotParsed<object>(TypeInfo.Create(err.Type), Enumerable.Empty<Error>());
             return err.Matched
-                ? AutoBuild(pr, current => DefaultParsingErrorsHandler(pr, current), e => e, maxDisplayWidth: maxDisplayWidth)
-                : AutoBuild(parserResult, current => DefaultParsingErrorsHandler(parserResult, current), e => e, true, maxDisplayWidth);
+                ? AutoBuild(pr, current => DefaultParsingErrorsHandler(pr, current), e => e, settings)
+                : AutoBuild(parserResult, current => DefaultParsingErrorsHandler(parserResult, current), e => e, settings, true);
         }
 
         /// <summary>
--- a/tests/CommandLine.Tests/Unit/ParserTests.cs
+++ b/tests/CommandLine.Tests/Unit/ParserTests.cs
@@ -860,5 +860,117 @@ namespace CommandLine.Tests.Unit
                     Assert.Equal("two", args.Arg2);
                 });
         }
+
+        [Fact]
+        // Tests a fix for issue #455
+        public void Help_screen_does_not_show_help_verb_if_AutoHelp_is_disabled()
+        {
+            var output = new StringWriter();
+            var sut = new Parser(config => { config.AutoHelp = false; config.HelpWriter = output; });
+
+            sut.ParseArguments<Secert_Verb, Add_Verb_With_Usage_Attribute>(new string[] { });
+
+            var helpText = output.ToString();
+            Assert.DoesNotContain("help", helpText, StringComparison.InvariantCulture);
+            Assert.DoesNotContain("Display more information on a specific command", helpText, StringComparison.InvariantCulture);
+        }
+
+        [Fact]
+        // Tests a fix for issue #455
+        public void Help_screen_does_not_show_help_option_if_AutoHelp_is_disabled()
+        {
+            var output = new StringWriter();
+            var sut = new Parser(config => { config.AutoHelp = false; config.HelpWriter = output; });
+
+            sut.ParseArguments<Simple_Options_With_HelpText_Set>(new string[] { });
+
+            var helpText = output.ToString();
+            Assert.DoesNotContain("--help", helpText, StringComparison.InvariantCulture);
+            Assert.DoesNotContain("Display this help screen.", helpText, StringComparison.InvariantCulture);
+        }
+
+        [Fact]
+        // Tests a fix for issue #455
+        public void Help_screen_shows_help_verb_if_AutoHelp_is_enabled()
+        {
+            var output = new StringWriter();
+            var sut = new Parser(config => { config.AutoHelp = true; config.HelpWriter = output; });
+
+            sut.ParseArguments<Secert_Verb, Add_Verb_With_Usage_Attribute>(new string[] { });
+
+            var helpText = output.ToString();
+            Assert.Contains("help", helpText, StringComparison.InvariantCulture);
+            Assert.Contains("Display more information on a specific command", helpText, StringComparison.InvariantCulture);
+        }
+
+        [Fact]
+        // Tests a fix for issue #455
+        public void Help_screen_shows_help_option_if_AutoHelp_is_enabled()
+        {
+            var output = new StringWriter();
+            var sut = new Parser(config => { config.AutoHelp = true; config.HelpWriter = output; });
+
+            sut.ParseArguments<Simple_Options_With_HelpText_Set>(new string[] { });
+
+            var helpText = output.ToString();
+            Assert.Contains("help", helpText, StringComparison.InvariantCulture);
+            Assert.Contains("Display this help screen.", helpText, StringComparison.InvariantCulture);
+        }
+
+        [Fact]
+        // Tests a fix for issue #414
+        public void Help_screen_does_not_show_version_verb_if_AutoVersion_is_disabled()
+        {
+            var output = new StringWriter();
+            var sut = new Parser(config => { config.AutoVersion = false; config.HelpWriter = output; });
+
+            sut.ParseArguments<Secert_Verb, Add_Verb_With_Usage_Attribute>(new string[] { });
+
+            var helpText = output.ToString();
+            Assert.DoesNotContain("version", helpText, StringComparison.InvariantCulture);
+            Assert.DoesNotContain("Display version information.", helpText, StringComparison.InvariantCulture);
+        }
+
+        [Fact]
+        // Tests a fix for issue #414
+        public void Help_screen_does_not_show_version_option_if_AutoVersion_is_disabled()
+        {
+            var output = new StringWriter();
+            var sut = new Parser(config => { config.AutoVersion = false; config.HelpWriter = output; });
+
+            sut.ParseArguments<Simple_Options_With_HelpText_Set>(new string[] { });
+
+            var helpText = output.ToString();
+            Assert.DoesNotContain("--version", helpText, StringComparison.InvariantCulture);
+            Assert.DoesNotContain("Display version information.", helpText, StringComparison.InvariantCulture);
+        }
+
+        [Fact]
+        // Tests a fix for issue #414
+        public void Help_screen_shows_version_verb_if_AutoVersion_is_enabled()
+        {
+            var output = new StringWriter();
+            var sut = new Parser(config => { config.AutoVersion = true; config.HelpWriter = output; });
+
+            sut.ParseArguments<Secert_Verb, Add_Verb_With_Usage_Attribute>(new string[] { });
+
+            var helpText = output.ToString();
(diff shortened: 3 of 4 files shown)


NeilMacMullen:
Looks like we were heading in similar directions. Still waiting to hear back on my PR https://github.com/commandlineparser/commandline/pull/458 but would be sensible to merge these at some point.

moh-hassan:
Thanks @x789  for your effort on this PR.
This help in fixing the issues: #414 and #455 without Breaking Change.

NeilMacMullen:
@moh-hassan 

******* PLEASE DO NOT MERGE ******

This change has been merged into #458

moh-hassan:
@x789 
I test the  custom help in v2.5.0 with  ` AutoVersion = false ` and `AutoHelp=false' and Version /Help are not displayed. 
This means that it works fine in v2.5.0 
Can you check this [online Demo ](https://dotnetfiddle.net/vk7WEO)

x789:
@moh-hassan,

setting AutoVersion and AutoHelp directly on HelpText works fine.

#414 and #455 describe a different issue: The settings of a Parser are 
not regarded when the help-text is automatically generated.
Thus, if you create a Parser and set its AutoHelp to false, it will be 
displayed on the auto-generated help-text. Please check out 
https://dotnetfiddle.net/UacitX for a demonstration of these issues. You 
will see that the ‘help’ option is still displayed if the help text is 
generated automatically (when using v2.5.0).

Best regards,
    x789

------ Originalnachricht ------
Von: "Mohamed Hassan" <notifications@github.com>
An: "commandlineparser/commandline" <commandline@noreply.github.com>
Cc: "x789" <x789@posteo.de>; "Mention" <mention@noreply.github.com>
Gesendet: 23.06.2019 23:45:38
Betreff: Re: [commandlineparser/commandline] Propagate parser-settings 
to HelpText (fix #414 and #455) (#461)

x789:
@moh-hassan,

setting `AutoVersion` and `AutoHelp` directly on `HelpText` works fine.

#414 and #455 describe a different issue: The settings of a `Parser` are not regarded when the help-text is automatically generated.
Thus, if you create a Parser and set its `AutoHelp` to false, it will be displayed on the auto-generated help-text. Please check out [this code](https://dotnetfiddle.net/UacitX) for a demonstration of these issues. You will see that the ‘help’ option is still displayed if the help text is generated automatically (when using v2.5.0).

moh-hassan:
@x789,  That is true as you said.

Based on your [example](https://dotnetfiddle.net/UacitX), It seems that in v2.4.3 and above, configuring parser with `AutoHelp = false`, cause other error and the help screen show twice `ERROR(S)` and it's supposed to show help without errors.

>ERROR(S):
  Required option 'i, input-file' is missing.

I [forked ](https://dotnetfiddle.net/LCNIFn) your test case to show 4 different behavior  

The valid behaviour is by avoiding configuring parser with `AutoHelp`

I think it's better to fix this bug and  unify the generation of help either from parser setting or help setting.
I found that you fix the bug by using an overloading of AutoBuild

            var settings = new ParserSettings() { AutoHelp = false, AutoVersion = false };
            var helpText = HelpText.AutoBuild(parserResult, settings);

Can you modify your PR to fix the bug described above and  unify the generation of help either from parser setting or help setting for all `AutoBuild` overloads.             

Note: , There are [18 test cases](https://github.com/commandlineparser/commandline/pull/462) for `HelpText` have been added.

NeilMacMullen:
@moh-hassan   @x789 This PR has already been merged into my PR #458 (as you requested).  Making further modifications on this branch seems a bit pointless assuming you are willing to accept my PR ! ;-)

moh-hassan:
@NeilMacMullen 
As your PR #458  is based on this PR, there is still a bug as described [above](https://github.com/commandlineparser/commandline/pull/461#issuecomment-505122066).
We can avoid the modification of PR #461 and concentrate for the fix in your PR.
Can you migrate that [example ](https://dotnetfiddle.net/vk7WEO) that configure both parser and HelpText by using your PR.
This can help developers to have a clean migration path from current help system to yours, specially most open issues are related to Help System.
As you know there was not a clean migration path from help system in v1.9.x to v2.x and we are in need to avoid this in the next V3.

NeilMacMullen:
@moh-hassan   that bug is not present on my branch.  I have added a test to verify that.

moh-hassan:
@NeilMacMullen , Good that the  bug is not present in your PR.
Just, i want to try the migration to your PR, can you migrate that [example](https://dotnetfiddle.net/vk7WEO) and post here as an attach file.
I use your new package in CI server to run the migrated code.

NeilMacMullen:
@moh-hassan 
Ok I think the 'new' version would look like this.  There's no need to call SetTextWriter if you're happy for HelpText to go to the console
```

                new Parser(c =>
                {
                    c.EnableDashDash = true;
                    c.CaseInsensitiveEnumValues = false;
                    c.IgnoreUnknownArguments = true;
                })
                .SetAutoHelp(false)
                .SetAutoVersion(false)
                .SetHelpTextConfiguration(c => { c.AdditionalNewLineAfterOption = false; })
                .ParseArguments<Options>(args)
                .WithParsed<Options>(options => Run(options));
```

moh-hassan:
Sweet Fluent interface Code :)
It's better for redirecting errors to Console.Error and others to Console.Out with user configuration. That  is one of the needs for most developers.

NeilMacMullen:
:-) Yes - you can just add a **.SetTextWriter(Console.Error)** to the chain if you want to do that

x789:
> ```
>                 new Parser(c =>
>                 {
>                     c.EnableDashDash = true;
>                     c.CaseInsensitiveEnumValues = false;
>                     c.IgnoreUnknownArguments = true;
>                 })
>                 .SetAutoHelp(false)
>                 .SetAutoVersion(false)
>                 .SetHelpTextConfiguration(c => { c.AdditionalNewLineAfterOption = false; })
>                 .ParseArguments<Options>(args)
>                 .WithParsed<Options>(options => Run(options));
> ```

@NeilMacMullen, please don't get me wrong. I admire your commitment to this project and your approach.

I personally like 'fluent' interfaces. But for me, the configuration quoted above is not symmetric. I'm afraid that users can get confused because they have to use two different types of configuration, an `Action<ParserSettings>` and methods like `SetAutoHelp`. What happens if I configure `Parser` to disable auto-help and configure `HelpText` to display auto-help?
> ```
>   new Parser(c => { c.AutoHelp = false })
>     .SetAutoHelp(true)
>     [...]
> ```

But maybe I have wrong expectations for the class `Parser`. When I use it, I don't want to have to deal with help text options. I expect an automatically generated help to take the parser settings into account without having to worry about it.

NeilMacMullen:
@789  Actually  I agree with youand appreciate the comments :-)  I had already decided to provide fluent modifiers for the parser settings. , not appreciating how they used up to now.  (One of the key changes in my PR is to differentiate between the settings used when _parsing_ and the setting used when generating the error text.  FWIW, I can't imagine myself every changing AutoHelp/Version but I _did_ want the error text to include the names of acceptable enum values which is currently quite clumsy using the old 'AutoBuild' version.


I'm not particularly happy that SetHelpTextConfiguration is required but that's an unfortunate consequence of "burying" a bunch of useful formatting flags in the HelpText class.  In the ideal world, these would be hoisted up and the Parser class would be the only thing that users would have to know about.  I'm open to ideas on better ways to expose these.

Your point about contradictory configurations is good but 1) it's also a problem (actually worse) in the original design and 2) only exists in the new design because of the desire to allow backwards compatibility.  I've actually marked all the old methods as Obsolete in my PR but would be quite happy to remove them and call this a breaking change.

moh-hassan:
@x789 , 
@NeilMacMullen
I did more investigation and found that the objective of `AutoHelp and AutoVersion` is different than show/hide objective.
>Added AutoHelp and AutoVersion properties to control adding of implicit 'help' and 'version' options/verbs

That is the original PR #256 with commit [5dad6e ](https://github.com/commandlineparser/commandline/tree/5dad6ed10ba50c2069cdecf8d621df5e9048bac9) and merged in v2.3.0 with commit [0db898d](https://github.com/commandlineparser/commandline/tree/0db898d8209955b112bf16f07e0e51e4373ad7f6)

It covers the issues #87 and #200 which are not related to issues #414 and #455 and the test cases used have different Assertion objective.

What are your suggestions ?

NeilMacMullen:
@moh-hassan Maybe I'm missing something but I don't think the behaviour obtained with the latest changes differs from that desired in #256?   In the latest changes, if you specify AutoHelp/AutoVersion=false then:

* the help/version verbs are not added to the parseable list and are treated as unrecognised items if the user tries to supply them
* help/version is not listed in the helptext.

Are you suggesting that something else should happen?

moh-hassan:
https://github.com/commandlineparser/commandline/pull/256#issuecomment-505529152

moh-hassan:
@NeilMacMullen 

- I want to confirm that we have a common understanding for the AutoHelp/AutoVersion feature and it should be used as given in PR #256.
- It's nice to get  the consideration mentioned  @x789 into account.
- The Fluent Interface API (FI) is a good enhancement and can be other API of using the library and it's best to be built  in the same common public API in `HelpText` without the need to be obsolete.  Any bugs in HelpText can be fixed in its place without the effect of FI.
- I try to understand the intent of the change in the issues #414 and #455 and I'll feedback.

moh-hassan:
There is a need for displaying help text with fluent syntax? #39

moh-hassan:
@x789 
@NeilMacMullen 

I have added a wiki documentation [page ](https://github.com/moh-hassan/commandline/wiki/Adding-of-implicit-Help-and-Version-screen-and-disabling-auto-generation) in my fork  to explain the usage of AutoHelp / AutoVersion properties of the parser with code example and four fiddle demos.
I hope It can help in understanding the PR #256

moh-hassan:
I think that #414 and #455 are now  a new feature of hiding Version/help option and are not considered as a bug.
Is't right?

x789:
#256 describes the possibility to automatically add a 'help' and 'version' verb/option to the parser. This works as far as only the parser is considered. I think when #256 was integrated, it was not considered that there will also be applications without 'help' and 'version' (see #294, #414 and #455).

@NeilMacMullen, @moh-hassan I do not want to stop the discussion about the fluent interface. But we should discuss it as part of #458. I am also confident that we will find a way to introduce a new API without breaking the old one.

@moh-hassan The wiki page looks okay to me. It describes the status quo. But it does not cover applications without 'help' and 'version' verbs/options.

I consider #414 and #455 to be bugs since the auto-generated help displays verbs/options that are not available. This is confusing and avoids that we can use the auto-help text in applications that do not have 'help' or 'version' verbs/options. If I interpret #294 correctly, others have come across this aspect as well.
v2.5.0 behaves as described in section 'Reproduction Steps' of #455:
>     2. Execute your program with `--help`
>        > ERROR(S):
>        > Verb '--help' is not recognized.
>        > --help       Display this help screen.
>        > --version    Display version information.

NeilMacMullen:
@x789  - yes I agree that most of this discussion should move to #458 - the only relevance here is that #458 aims to remove the requirement for the user ever to use or know about AutoBuild and therefore uses your changes in a slightly modified version.  (Hence I'm keen to know which direction we are going with this - if this PR is merged then I'll just end up having to redo changes.)

@moh-hassan 
I'm pretty sure that both this PR and #458 are correctly interpreting the original intention behind introducing the AutoHelp and AutoVersion  flags which was to treat these as flags which suppress the automatic generation of the associated commands and options.  The hypothetical user of AutoHelp=false would be expected to 
1) Supply their own "help" verb (assuming the wanted one)
2) Suppress the standard AutoBuild "verb not recognised" command by setting TextWriter to null and/or using WithNotParsed to output their own first level help screen.

On you wiki you seem to be implying that the suppression of AutoBuild output might happen automatically if AutoHelp was set to false.  Personally I think that is a step too far and undesirable for users who want the automatically generated basic usage but who want to supply their own "help" behaviour.

I tried your Fiddle and couldn't get to output anything sensible.    For example, I commented out the "with.AutoVersion=false;" line and the WithNotParsed line and got absolutely no output.  Is that based on the current state of the develop branch?

NeilMacMullen:
@moh-hassan   Played some more with your Fiddle and I think 2.5 is broken in some ways that have nothing to do with AutoHelp…

1) https://dotnetfiddle.net/uTUhKc - dual error line (nothing to do with AutoHelp)
2) https://dotnetfiddle.net/BZgB4M - missing autobuild output when not setting config flags
3) https://dotnetfiddle.net/gz1Rej - missing autobuild output even when AutoHelp set to true explicitly

It looks more like the explicit Parser configuration mechanism is broken in this release so I'm not sure the current behaviour is really a good guide for how it _should_ work.  If I had to guess, I would assume that when calling the constructor with a no-op function, the ParserSettings object is a new one with a null TextWriter rather than the settings used by Parser.Default.  This seems like a very bad disconnect - **Parser.Default** should give equivalent behavior to **new Parser(_=>{})**.

moh-hassan:
@NeilMacMullen
>I think 2.5 is broken in some ways that have nothing to do with AutoHelp…


The  AutoHelp/AutoVersion feature  is added  in v2.4.3 as described [here](https://github.com/commandlineparser/commandline/pull/461#issuecomment-505517415) and no break in v2.5.
You can try v2.4.3 and you get the same result.
Can you point me to a commit that may cause a break of that feature.


The feature is working as intended and tested based on the requirement of @nemec (one of the Authors of the library)

 @nemec [commented ](https://github.com/commandlineparser/commandline/pull/256#issuecomment-376300704)on Mar 26, 2018


>I think this solves the issues nicely. Can you add a few tests to help us prevent regressions? I think the >following should cover things, but if you think of any other scenarios feel free to include them:
>
>    AutoHelp = false and invoking --help returns an error.
>    AutoVersion = false and invoking --version returns an error.
>    AutoHelp = false, a flag for --help added manually by the user to the Options class, and it can be invoked successfully.
>    AutoVersion = false, a flag for --version added manually by the user to the Options class, and it can be invoked successfully.

The tests in the original commit [here ](https://github.com/commandlineparser/commandline/pull/256/files#diff-5ccde8f7110f8ef6f87c2a1397757c83R1018) and in v2.5.0 [here](https://github.com/commandlineparser/commandline/blob/4d69a3898b84c821161b2db948075cd9e162832c/tests/CommandLine.Tests/Unit/Core/InstanceBuilderTests.cs#L1052)

For fiddle, there is a problem in redirection of  console.Error and duplicate the errors
I run your fiddle cases in Linqpad  and there is no duplication of errors.
For custom Parser, HelpWriter = null by default so you get no output, this is working

                var parser = new CommandLine.Parser(c=>{c.HelpWriter=Console.Out;});

So AutoHelp/AutoVersion is working as intended and approved by the @nemec.
If something wrong in this feature we can discuss it and provide a solution.

NeilMacMullen:
@moh-hassan I must admit I am very confused ;-) You were the one who suggested using Fiddle to try and explore the behaviour, with the implication that this PR (and therefore mine) were somehow getting it wrong.  Since  regression tests are passing on both PRs I'm not really sure what you think should be changed!  Perhaps if you could point to a failing test case it would be clearer?  (Maybe there is one in an unmerged PR?)  Otherwise we just seem to be going round in circles....  

The fact that the "default" new Parser constructor behaviour is different from Parser.Default is confusing and something I think should be considered a bug (or at least a poor design decision) but is not directly relevant to how AutoHelp should behave except in as much that it just reveals that the current implementation has too many inconsistencies and it's therefore difficult to use it to establish the "desired" behaviour.   I still think the actual behaviour of this PR (and #458) is sensible and desired.  Can you explain where you think it isn't?

x789:
> @nemec [commented ](https://github.com/commandlineparser/commandline/pull/256#issuecomment-376300704)on Mar 26, 2018
> 
> > I think this solves the issues nicely. Can you add a few tests to help us prevent regressions? I think the >following should cover things, but if you think of any other scenarios feel free to include them:
> > AutoHelp = false and invoking --help returns an error.
> > AutoVersion = false and invoking --version returns an error.
> > AutoHelp = false, a flag for --help added manually by the user to the Options class, and it can be invoked successfully.
> > AutoVersion = false, a flag for --version added manually by the user to the Options class, and it can be invoked successfully.

The requirements stated above are correctly implemented in CommandLineParser v2.5.0. _But_ in my (and others, see #294 and #414) opinion two important scenarios are missing:
- AutoHelp = false and there is no manually added verb/option called 'help'
- AutoVersion = false and there is no manually added verb/option called 'version'

I suspect that if these requirements had been considered in #256, it would have been noticed that the automatically generated `HelpText` does not match the parser configuration.

NeilMacMullen:
>two important scenarios are missing

Indeed, and worth adding there are now unit tests that explicitly check these scenarios.... [see line 681](https://github.com/commandlineparser/commandline/pull/461/files#diff-98f09089b58f94c3bad853ffe0513a47) albeit together.  In #458 I've broken the combined test into two individual ones for paranoia's sake, but the behaviour is the same. See [line 678](https://github.com/commandlineparser/commandline/pull/458/files#diff-9ab5b530e14f7ff6d1e8ddb9bd8018fd)

moh-hassan:
@NeilMacMullen 
I use fiddle as a way of simple quick online demenstoration even with  this small bug of redirection :)
For deep debug I go to Vs2017 or Linqpad.


My concerns is that the  intention of the original issues #87 and #200 (disable help/version generation by building their own custom HelpText) which is fixed by Pr #256 are  different than the intention of the issues #414 and #455 (which simply hide help/version option from auto generated help) and therefore should have different resolution.

BTW, --version and --help are  LNX/POSIX standard in all commandLine utilities (this library respect and implement this standard ) and the requirement of the issues #414 and #455 for hiding them may violate this standard.
Are we ready to violate this standard?!

Any enhancement or Fluent interface API to HelpText is welcome and it's needed in #39 and can be completed with this  PR and yours  without Breaking Change or obsolete AutoBuild.
AutoBuild is the engine of generating help for Parser.Default.

x789:
@moh-hassan
> My concerns is that the intention of the original issues #87 and #200 (disable help/version generation by building their own custom HelpText) which is fixed by Pr #256 are different than the intention of the issues #414 and #455 (which simply hide help/version option from auto generated help) and therefore should have different resolution.

I think this is a misunderstanding. This pull request #461 has nothing to do with the configuration of `HelpText`. It simply hides 'help' and 'version' if these verbs/options are not understood by the current parser configuration. Nothing else.

> BTW, --version and --help are LNX/POSIX standard in all commandLine utilities (this library respect and implement this standard ) and the requirement of the issues #414 and #455 for hiding them may violate this standard.
> Are we ready to violate this standard?!

Unfortunately, CommandLineParser violates the POSIX standard by allowing multi-character options, double-dashes and verbs. [Chapter 12.2 is very precise.](http://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap12.html) 


I wasn't aware that CommandLineParser 'enforces' the presence of help and version. If this restriction is really wanted, it should be documented in a prominent place.

moh-hassan:
@x789 
I was evaluating CommandLineParser against POSIX standard in this [comment](https://github.com/commandlineparser/commandline/issues/357#issuecomment-435513063)

The Readme link this [Getopt ref](https://www.gnu.org/software/libc/manual/html_node/Getopt.html)

Can you give me an example of the violation of the multi-character options?

>If this restriction is really wanted, it should be documented in a prominent place.

Yes, you are correct. Attention will be take for the documentation and I hope contribution in Wiki.

moh-hassan:
@x789 
> two important scenarios are missing:
    AutoHelp = false and there is no manually added verb/option called 'help'
    AutoVersion = false and there is no manually added verb/option called 'version'

These two scenarios are described in the [wiki ](https://github.com/moh-hassan/commandline/wiki/Adding-of-implicit-Help-and-Version-screen-and-disabling-auto-generation#example2-disabling-autoversion) in 
[Example1: Disabling AutoHelp](https://github.com/moh-hassan/commandline/wiki/Adding-of-implicit-Help-and-Version-screen-and-disabling-auto-generation#example1-disabling-autohelp)
[Example2: Disabling Version](https://github.com/moh-hassan/commandline/wiki/Adding-of-implicit-Help-and-Version-screen-and-disabling-auto-generation#example2-disabling-autoversion)

In these scenarios, parser consider --help and --version an unknown option and raise an error
and consequently no help/version screen are generated.

Developers can handle these cases (if needed) in:

                     parserResult.WithNotParsed(ShowError);

and show their own custom help/version or custom error screen

x789:
@moh-hassan 
You type faster than I do :-). I hope to catch up...

> I was evaluating CommandLineParser against POSIX standard in this [comment](https://github.com/commandlineparser/commandline/issues/357#issuecomment-435513063)
> 
> The Readme link this [Getopt ref](https://www.gnu.org/software/libc/manual/html_node/Getopt.html)

Thank you for referring to the getopt-documentation. I can understand your point of view now.
However, I think that a 'long option' is against the POSIX standard. According to POSIX, an option may only be designated by a single character. (This is the first item on the list.) In contrast, the 'GNU option characteristics' allow multiple characters.

~~But for me the crucial question is whether CommandLineParser is based on the constraint that a program must always have 'help' and 'version'. If it does, then #414 and #455 are no bugs and the auto-generated `HelpText` is fine. I would then treat them as feature requests which I would reject with reference to that constraint.~~ (see my comment below)

x789:
With the knowledge gathered in this discussion, I have once again read through the complete documentation of the module.

@moh-hassan I can now completely understand your argumentation. Some points are hidden between the lines in the documentation. In my experience, however, it is normal for a developer documentation not to contain and explain every (implicit) decision made.

Here's my suggestion on how to proceed:
- This PR and Issues #294, #414 and #455 are closed. This is because the automatisms in `Parser` assume that the verbs/options 'help' and 'version' are present. (Regardless of whether they are created by the user or by the parser.)
- @moh-hassan's new wiki page "[Adding of implicit Help and Version screen and disabling auto generation](https://github.com/moh-hassan/commandline/wiki/Adding-of-implicit-Help-and-Version-screen-and-disabling-auto-generation#example2-disabling-autoversion)" will be added to the official wiki.
    - I suggest that the first note be changed to "This feature control the generation of version/help _verbs/options_ and it is not hiding help/version options in the auto generated help."
- The reference to GNU getopt is more prominent in the ReadMe. It is moved from "Resources for NewComers" to "At a Glance".
- We create a new feature request for `HelpText.AutoBuild`. This describes that you can specify whether 'help' or 'version' should be displayed. You can set this property for instances of `HelpText`, but not when calling `AutoBuild()`.

moh-hassan:
@x789 
Thanks for your suggestions.
I'll convert your words into action plan.
The  feature request for HelpText.AutoBuild to hide/show help/version option  as a property in the HelpText is the exact solution.

moh-hassan:
Closed based on @x789 in his [comment](https://github.com/commandlineparser/commandline/pull/461#issuecomment-506237242)

Athari:
@moh-hassan

> My concerns is that the intention of the original issues #87 and #200 (disable help/version generation by building their own custom HelpText) which is fixed by Pr #256 are different than the intention of the issues #414 and #455 (which simply hide help/version option from auto generated help) and therefore should have different resolution.

#414 and #455 don't look like "simply hiding" for me. Hiding an unavailable command is correct. Hiding an available command is a bug. So the main problem is that the pull request #256 didn't test generation of error message help properly.

@x789's suggestion of completely separating availability as a command and availability on the help screen sounds really weird to me, to be frank. Maybe I'm missing something or I'm unaware of technical limitations, but that's not what users of command line programs would *ever* expect.

> BTW, --version and --help are LNX/POSIX standard in all commandLine utilities (this library respect and implement this standard ) and the requirement of the issues #414 and #455 for hiding them may violate this standard.
> Are we ready to violate this standard?!

Removing "version" and "help" completely is bad, but I think it should be allowed. The library is not standards police.

> These two scenarios are described in the wiki

I can't access wiki for some reason, GitHub keeps redirecting me to the front page of the fork.

PR #464, opened 2019-06-27
Add examples that I wish were there

ben-biddington:
I wish this style of usage example was already around somewhere, or easier to find if they are!

These ones show how to:

* parse options and get result without having to use `WithParsed`
* make options that accept multiple values.

I tried to make these as clear as possible, in the hope they may be understood by reading just the one file.

I realise there is duplication here, but perhaps it is worth it in cases like this. (The duplication is bound to public interface members that are very stable.)


ChrisUnityArto:
This was very helpful to me - it seems this multiple values feature is undocumented apart from this example, so I think it should be merged. Thanks @ben-biddington !

Issue #259, opened 2018-03-21
How can I use a custom HelpText for help/error output?


tmillican:
More particularly, I'm trying to force a line break spacer between the copyright notice and the "USAGE:" line (and between "USAGE:" and "ERROR(S):". But in the general case, I can't figure out how to inject a custom `HelpText` instance into the output process. I've read [#224](https://github.com/commandlineparser/commandline/issues/224), but this approach doesn't seem to actually work -- at least not in 2.2.1.

Example code:

```
namespace ClpTest
{
    using System;
    using System.Collections.Generic;
    using CommandLine;
    using CommandLine.Text;

    internal class Options
    {
        [Option('f', "foo", Required = true, HelpText = "Uses the specified foo.")]
        public string Foo { get; }

        [Option('b', "bar", Required = true, HelpText = "Uses the specified bar.")]
        public string Bar { get; }

        public Options(string foo, string bar)
        {
            Foo = foo;
            Bar = bar;
        }
    }

    class Program
    {
        static void Main(string[] args)
        {
            var parser = Parser.Default;
            var result = parser.ParseArguments<Options>(args)
                .WithParsed<Options>(opts => DoParsed(opts))
                .WithNotParsed(errs => DoError(errs));
        }

        internal static void DoParsed(Options opts)
        {
            Console.WriteLine($"DoParsed invoked with Foo={opts.Foo}, Bar={opts.Bar}.");
        }

        internal static void DoError(IEnumerable<Error> errs)
        {
            Console.WriteLine($"DoError() invoked.");
        }
    }
}
```

If you run this, you'll see that `DoError(...)` isn't invoked until _after_ the help text is output. If you define a custom HelpText in `DoError(...)` and output it there, as suggested in #242 and elsewhere, you can get whatever output you'd like, but only after the default help text has already been printed.

After poking at the source code, it seems that `Parser.ParseArguments<T>(...)` invokes `Parser.MakeParserResult<T>(...)`, which invokes `Parser.DisplayHelp<T>(...)`, which ultimately invokes `HelpText.AutoBuild<<T>(...)`. The only control the user has over this process (that I can determine) is the ParserSettings object used to construct the Parser. This lets you specify which TextWriter to use, and whether to print line breaks between options, but that's all.

As a workaround, I can specify a `TextWriter` object that blackholes the output and then substitute my own output in `DoError(...)`, but it's pretty kludgey.

[...]

PR #467, opened 2019-07-03
Fix issues  for HelpText.AutoBuild configuration (issues #224 , # 259)

moh-hassan:
HelpText.AutoBuild can't apply setting on HelpText to add AdditionalNewLineAfterOption and other setting. 
These setting are applied only when parser raise parsing errors.
This PR fix this bug and fix issue #224 , #259 .
You have a complete control on HelpText configuration in custom help.
Also, new extension methods IsHelp() /IsVersion() is added to simplify checking errors if it have help/version option.

            var parser = new Parser(x =>
            {
                x.HelpWriter = null;               
            });
            var result = parser.ParseArguments<Simple_Options>(new[]{"--help"});
            //generate custom help 
            result .WithNotParsed(errs =>
            {               
                var helpText = HelpText.AutoBuild(result,
                    h =>
                    {
					    //configure help
                        h.AdditionalNewLineAfterOption = false;
                        return h;
                    }
                    , e => e);
                 //print help screen
                Console.WriteLine(helpText);
                 
            });

### DIFF
--- a/src/CommandLine/Text/HelpText.cs
+++ b/src/CommandLine/Text/HelpText.cs
@@ -269,11 +269,13 @@ namespace CommandLine.Text
 
             var errors = Enumerable.Empty<Error>();
 
+         
             if (onError != null && parserResult.Tag == ParserResultType.NotParsed)
             {
                 errors = ((NotParsed<T>)parserResult).Errors;
-
-                if (errors.OnlyMeaningfulOnes().Any())
+                if (errors.IsHelp())
+                    auto = onError(auto);
+                else if (errors.OnlyMeaningfulOnes().Any())
                     auto = onError(auto);
             }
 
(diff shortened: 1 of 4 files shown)


src/CommandLine/Text/HelpText.cs
```diff
@@ -269,11 +269,13 @@ public SentenceBuilder SentenceBuilder
...
 
+         
             if (onError != null && parserResult.Tag == ParserResultType.NotParsed)
             {
                 errors = ((NotParsed<T>)parserResult).Errors;
-
-                if (errors.OnlyMeaningfulOnes().Any())
+                if (errors.IsHelp())
+                    auto = onError(auto);
+                else if (errors.OnlyMeaningfulOnes().Any())
```
asherber: I suggest simplifying: 
```csharp
if (errors.IsHelp() || errors.OnlyMeaningfulOnes().Any())
   auto = onError(auto);

moh-hassan: Thanks for your suggestions.
I'll modify my code to use yours.

moh-hassan:
This PR also fix #414 and #455

PR #476, opened 2019-07-16
Fix a simple typo in SentenceBuilder

skylarnam:
Just a simple typo fix.


ajtruckle (approved):
seems ok

ajtruckle:
This is 4 years old. I can only assume that none of the collaborators are maintaining it any more.

PR #478, opened 2019-07-16
Allow BaseAttribute HelpText localization to use internal classes and properties.

### LINKED ISSUE #477: BaseAttribute HelpText localization can not use internal properties.
PR #356 fixes issue #330 by allowing HelpText to be read from a static property. However, the current implentation restricts the use of internal [classes](https://github.com/tkouba/commandline/blob/0881d1b7e1277859a67fe93350f907c9ec377119/src/CommandLine/Infrastructure/LocalizableAttributeProperty.cs#L47) and [properties](https://github.com/tkouba/commandline/blob/0881d1b7e1277859a67fe93350f907c9ec377119/src/CommandLine/Infrastructure/LocalizableAttributeProperty.cs#L53). 

Resx files are a common way to store localized text, and MSBuild will automatically generate a class with static properites representing the localized value, by using the resgen utility. By default, [resgen will generate an internal class with internal properties](https://docs.microsoft.com/en-us/dotnet/framework/tools/resgen-exe-resource-file-generator#generating-a-strongly-typed-resource-class). 

I propose the visibility restrictions be relaxed, as the reflection code being called works regardless of visibility.

ILMTitan:
Fixes #477.

### DIFF
--- a/src/CommandLine/Infrastructure/LocalizableAttributeProperty.cs
+++ b/src/CommandLine/Infrastructure/LocalizableAttributeProperty.cs
@@ -43,10 +43,9 @@ namespace CommandLine.Infrastructure
                 return _value;
             if (_localizationPropertyInfo == null)
             {
-                // Static class IsAbstract 
-                if (!_type.IsVisible)
-                    throw new ArgumentException("Invalid resource type", _propertyName);
-                PropertyInfo propertyInfo = _type.GetProperty(_value, BindingFlags.Public | BindingFlags.GetProperty | BindingFlags.Static);
+                PropertyInfo propertyInfo = _type.GetProperty(
+                    _value,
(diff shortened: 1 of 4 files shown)


moh-hassan:
Thanks @ILMTitan  for your effort.

Localization in .NET is based on public (not internal) resources. For example, in `DataAnnotation DisplayAttribute`, it  check for visibility  and  fire exception for internal resources .

because:
- Resources may move to satellite assemblies, and internal resources are not visible to other assemblies.
- In WPF application, Xaml can't use internal resources.

For these reasons, localization in PR #356 check for visibility and fire exception for internal resources .

I know, in VS IDE by default, resources are internal, but this can be modified easily.

Is there any problems in using public resources?

ILMTitan:
Moving discussion to issue #477.

Issue #482, opened 2019-07-19
Order of option order in help text


tehkyle:
Appologies if this is already possible or answered somewhere, I couldn't find details in the docs or searching existing issues.

Could the Help Text generator take an order for the listed options, such as by Required first, then Alphabetical?

We have a set of available verbs that have some shared options so have built our option classes to be inherited. Currently the help builder lists these options in inherited order, but we would like to at least collect required options at the top.

moh-hassan:
The feature is not available.
PR is welcome.

b3b00:
I take the issue. PR will come in some days may be hours.
So the solution will be : order option by

1. required than optional
2. alphabetical order.

Would you like a way (C# attribute or some param on HelpTExt.AutoBuild(...)) to enable it ?
I would go for an lambda that specify the desired order with a static lambdas for

- actual ordering
- your desired ordering (required than alpha)
- maybe others like just alpha

b3b00:
I made some dev and finally reworked it.
Coming PR allow to redefine order using a ```Comparison<ComparableOption>``` function. 
I choose to create a ComparableOption from the Specification class to not break existing visibility rules.
To redefine order you then have to provide de Comparison object as a static value of HelpText class before calling AutoBuild :

```c#
HelpText.OptionComparison = HelpText.RequiredThenAlphaComparison;
var helpMessage = HelpText.AutoBuild(parseResult,...) 
```

Maybe not the perfect solution but it works well.
PR comes with the  order that @tehkyle proposed.
When no comparison is provided before calling AutoBuild then it defaults to the current behavior (declaration order)

PR #484, opened 2019-07-23
issue #482: reorder options in auto help text

b3b00:
this PR allows to redefine order using a Comparison<ComparableOption> function.
I choose to create a new ComparableOption struct from the Specification class to not break existing visibility rules (Specification is internal).
To redefine order you then have to provide a Comparison object as a static property of HelpText class before calling AutoBuild :

```c#
HelpText.OptionComparison = HelpText.RequiredThenAlphaComparison;
var helpMessage = HelpText.AutoBuild(parseResult,...) 
```

Maybe not the perfect solution but it works well.
PR comes with the order that @tehkyle proposed.
When no comparison is provided before calling AutoBuild then it defaults to the current behavior (declaration order)

Unit test is also provided

Maybe a better and more elegant solution would be to provide a fluent API to configure HelpText AutoBuild.

Something like 
```c#
HelpText.With(parseResult)
                .OnError((err) => { })
                .OnExample(example) => { })
                .WithOrder((ComparableOption attr1, ComparableOption attr2) => 1)
                .Build();
```

@maintaners what do you think ? do you prefer I rework it with fluent API or can we stay with it ?


b3b00:
I 've finally moved to a fluent API design as I found static properties are hell.
I kept the existing AutoBuild method not to break existing code.

```c#
var message = HelpText.CreateWith(parseResult)
                .WithComparison(HelpText.RequiredThenAlphaComparison)
                .OnError(error => {
                    throw new InvalidOperationException($"help text build failed. {error.ToString()}");
                })
                .OnExample(ex =>
                    {
                        return null;
                    })
                .Build();
```

moh-hassan:
Thanks  Olivier Duhart  for your effort.
Your PR cover two valuable features: Reorder options and Fluent Interface.
It's preferred that every feature has its own branch/PR so every PR pass its own test locally and CI server then merged.
Can you move the  Fluent Interface feature to a separate branch/PR and let this PR for Reorder options only and pass CI checks.

b3b00:
@moh-hassan , CI fails on a sensless unit test. So wouldnt it be easier to just remove this unit test and let CI succeed ?
I tried to order option on 

1. required option
2. alphabetical order on option short name

but implicit options (--help and --version) does not have short name and then appeared before "real" options with my naive implementation of the comparison function. My bad I did not even try my UT before pushing it ! shame on me.

I have a local version of the UT that is OK : I 've simply modify the comparison function to take account of this particular edge case.

moh-hassan:
The fail at [line 186](https://github.com/commandlineparser/commandline/blob/04a9bcf59633b644d1df5614ca7e7e90ead7f3a3/tests/CommandLine.Tests/Unit/Issue482Tests.cs#L186)
It seems that you test the ordering using the new syntax Fluent Interface (FI).
Try to  do your unit test of ordering options using the standard AutoBuild method and Isolate FI.
That is why it's necessary to  move the Fluent Interface feature to other PR with its own unit tests.

moh-hassan:
Good that you discovered the fail reason. 
Test can cover: no-short-name/no-long-name/no-short-and-no-long-name/both-short-and-long-name.

moh-hassan:
you can benefit from [NameText](https://github.com/commandlineparser/commandline/blob/master/src/CommandLine/NameInfo.cs#L49) : Gets a formatted text with unified name information.

b3b00:
@moh-hassan concerning unit tests, as order is defined using a custom Comparison<T> function, I simply test with 3 cases : 

- order by required then alpha longName as asked by @tehkyle in issue.
- order by required then alpha shortName : this one for testing comparison overriding
- no comparison : this one to that behavior defaults to previous behavior (ordered by declaration order)

More tests are not that usefull I guess

moh-hassan:
@Olivier 
HelpText fluent API shouldn't be part of this PR and I suggest moving it to other PR.

b3b00:
I 've reverted all fluent API code. I 'll come back with the fluent API enhancement later if PR accepted and merged in develop.
So the use case is now : 

```c#
Comparison<ComparableOption> myComparison= (ComparableOption attr1, ComparableOption attr2) => {
// some sensible comparison code in here
return 1; }
string message = HelpText.AutoBuild(
    parseResult, 
    error => {return null;},
    ex => { return null;},
    comparison:comparison
);
```

src/CommandLine/Text/HelpText.cs
```diff
@@ -17,8 +17,71 @@ namespace CommandLine.Text
...
+    
     public class HelpText
     {
+
+
+        ComparableOption ToComparableOption(Specification spec, int index)
+        {
+            OptionSpecification option = spec as OptionSpecification;
+            ValueSpecification value = spec as ValueSpecification;
+            bool required = option != null ? option.Required : false;
```
moh-hassan: Simplify to:

               bool required = option?.Required ?? false;

src/CommandLine/Text/HelpText.cs
```diff
@@ -17,8 +17,71 @@ namespace CommandLine.Text
...
+                LongName = option?.LongName,
+                ShortName = option?.ShortName,
+                Index = index
+            };
+        }
+
+
+        public Comparison<ComparableOption> OptionComparison = null;
+
+        public static Comparison<ComparableOption> RequiredThenAlphaComparison  = (ComparableOption attr1, ComparableOption attr2) =>
```
moh-hassan, line 58: Remove redundant else and braces, and use StringComparison.Ordinal

           public static Comparison<ComparableOption> RequiredThenAlphaComparison  = (attr1, attr2) =>
        {
            if (attr1.IsOption && attr2.IsOption)
            {
                if (attr1.Required && !attr2.Required)
                    return -1;

                if (!attr1.Required && attr2.Required)
                    return 1;

                int t = string.Compare(attr1.LongName, attr2.LongName, StringComparison.Ordinal);
                return t;
            }

            return attr1.IsOption && attr2.IsValue ? -1 : 1;
        };

src/CommandLine/Text/HelpText.cs
```diff
@@ -17,8 +17,71 @@ namespace CommandLine.Text
...
+        public bool IsValue;
+        public string LongName;
+        public string ShortName;
+        public int Index;
+    }
+    
     public class HelpText
     {
+
+
```
moh-hassan, line 37: #region Option ordering

src/CommandLine/Text/HelpText.cs
```diff
@@ -17,8 +17,71 @@ namespace CommandLine.Text
...
+            else if (attr1.IsOption && attr2.IsValue)
+            {
+                return -1;
+            }
+            else
+            {
+                return 1;
+            }
+        };
+
```
moh-hassan, line 83: #endregion

tests/CommandLine.Tests/Unit/Issue482Tests.cs
```diff
@@ -0,0 +1,200 @@
...
+            string expectedCompany = "Company";
+
+
+            var parser = Parser.Default;
+            var parseResult = parser.ParseArguments<Options_HelpText_Ordering_Verb1, Options_HelpText_Ordering_Verb2>(
+                    new[] { "verb1", "--alpha", "alpaga", "--alpha2", "alala", "--charlie", "charlot" })
+                .WithNotParsed(errors => { throw new InvalidOperationException("Must be parsed."); })
+                .WithParsed(args => {; });
+
+            Comparison<ComparableOption> orderOnShortName = (ComparableOption attr1, ComparableOption attr2) =>
```
moh-hassan, line 121: Remove redundant else and braces, and use StringComparison.Ordinal as done [here](https://github.com/commandlineparser/commandline/pull/484/files#r306971344)

moh-hassan:
It looks fine. 
-	You add a new parameter with default null to AutoBuild method. That is working.
To minimize the parameters passed to  AutoBuild method, What about if it's set   as a public property like other setting in HelpText :
	 
        public Comparison<ComparableOption> Comparison {get;set;}

- Provide test case with some properties as value options one required and the other Value properties not required.
- For documentation, Can you provide a wiki page describing how to to use the Order option feature with examples.

b3b00:
@moh-hassan 
Regarding your remark on the Comparison value as an instance attribute : I agree there is to much parameters. 
The autobuild method build the  HelpText instance so we have to give it the optional comparison object if we want it to be used. 
At the end of the method HelpText object is built AND helpmessage computed.
 I don't see any simple method to remove parameters still keeping all the possibilities.
that was one of the purpose of my fluent API proposal : make it clearer to build an HelpText and compute message.

About your test case asking : my test case is the one you describe

- 2 required options
- 3 not required options
- 1 value

Could you give an example of a test case, please ?

About the wiki, I will do it when all dev done and checked, and when time available .

-

moh-hassan:
All help configuration is supposed to be done using `error Action`, but this configuration is effective only (due to a bug in `HelpText`) when the parser fire parsing errors and it's not effective when using "--help".
This bug is resolved in the PR  #467  (it's not merged yet) and now all configuration can be done using `error Action`
I merged your PR with PR #467 locally and did POC to use an instance variable instead of a parameter and it's working fine.

         var   message = HelpText.AutoBuild(parseResult, error =>
            {
                error.OptionComparison = HelpText.RequiredThenAlphaComparison;                
		return error;				
            },
            ex =>ex );   
I can merge PR #467  to develop.
What are your suggestions?

b3b00:
@moh-hassan I would be easier if you merge #467 to develop, then I will rebase my branch with develop and do the fix according to #467 fix. 
It look more appealing than the way I've done it : too many parameters I agree.
Let me know if you agree and do the merge.

thanks 

Olivier

moh-hassan:
It's merged.
Now test case can be:

         [Fact]
        public void AutoBuild_with_ordering()
        {      
            var parser = Parser.Default;
            var parseResult = parser.ParseArguments<Options_HelpText_Ordering_Verb1, Options_HelpText_Ordering_Verb2>(
                    new[] { "verb1", "--help" }) //call help
                .WithNotParsed(errors => { ; })
                .WithParsed(args => {; });         

           //AutoBuild using instance variable without passing comparison  parameter 
            string message = HelpText.AutoBuild(parseResult, 
                error => 
				{
					 error.OptionComparison = HelpText.RequiredThenAlphaComparison;  
					return error;
				},
                ex => ex);
                   
			var helps = message.ToString().ToNotEmptyLines().Skip(2).ToList();
           var expected = new List<string>()
            {
                "  -a, --alpha      Required.",
                "  -b, --alpha2     Required.",
                "  -c, --bravo",
                "  -d, --charlie",
                "-e, --echo",
                "-f, --foxtrot",
                "--help           Display this help screen.",
                "--version        Display version information.",
                "value pos. 0"
            };
            Assert.Equal(expected.Count, helps.Count);
            int i = 0;
            foreach (var expect in expected)
            {
                Assert.Equal(expect.Trim(), helps[i].Trim());
                i++;
            }           
        }

src/CommandLine/Text/HelpText.cs
```diff
@@ -240,21 +304,24 @@ public SentenceBuilder SentenceBuilder
...
+        /// <param name="comparison">a comparison lambda to order options in help text</param>
         /// <remarks>The parameter <paramref name="verbsIndex"/> is not ontly a metter of formatting, it controls whether to handle verbs or options.</remarks>
         public static HelpText AutoBuild<T>(
             ParserResult<T> parserResult,
             Func<HelpText, HelpText> onError,
             Func<Example, Example> onExample,
             bool verbsIndex = false,
-            int maxDisplayWidth = DefaultMaximumLength)
+            int maxDisplayWidth = DefaultMaximumLength,
+            Comparison<ComparableOption> comparison = null)
```
moh-hassan: //remove this line, no more parameters are used
//Comparison<ComparableOption> comparison = null

src/CommandLine/Text/HelpText.cs
```diff
@@ -240,21 +304,24 @@ public SentenceBuilder SentenceBuilder
...
         {
             var auto = new HelpText
             {
                 Heading = HeadingInfo.Empty,
                 Copyright = CopyrightInfo.Empty,
                 AdditionalNewLineAfterOption = true,
                 AddDashesToOption = !verbsIndex,
-                MaximumDisplayWidth = maxDisplayWidth
+                MaximumDisplayWidth = maxDisplayWidth,
+                OptionComparison = comparison
```
moh-hassan: //remove this line, no more parameters are used
// OptionComparison = comparison

tests/CommandLine.Tests/Unit/Issue482Tests.cs
```diff
@@ -0,0 +1,197 @@
...
+            Comparison<ComparableOption> comparison = HelpText.RequiredThenAlphaComparison;
+
+            string message = HelpText.AutoBuild(parseResult,
+                error =>
+                {
+                    error.OptionComparison = HelpText.RequiredThenAlphaComparison;
+                    return error;
+                },
+                ex => ex,
+                comparison: comparison);
```
moh-hassan: //remove this line
//comparison: comparison

src/CommandLine/Text/HelpText.cs
b3b00: working on it.

tests/CommandLine.Tests/Unit/Issue482Tests.cs
```diff
@@ -0,0 +1,197 @@
...
+            string message = HelpText.AutoBuild(parseResult,
+                    error =>
+                    {
+                        error.OptionComparison = orderOnShortName;
+                        return error;
+                    },
+                    ex => ex,
+                        false,
+                        80,
+                        orderOnShortName);
```
moh-hassan: //remove the line orderOnShortName
//Enough to call without extra parameters:

            string message = HelpText.AutoBuild(parseResult,
                    error =>
                    {
                        error.OptionComparison = orderOnShortName;
                        return error;
                    },
                    ex => ex);

b3b00:
I have trouble with  onerror action. I sync my fork with upstream , I see your commandline#439 merge but it seems it does not call onerror when parse is ok.
I look at the AutoBuild code and indeed if parse is ok it does not call onerror. did something go wrong when merging ? could you check ?

moh-hassan:
> indeed if parse is ok it does not call onerror

That is correct.
onerror is called when parser fire parsing error or call '--help'
You need to modify the test cases and call help as given below ( my [example ](https://github.com/commandlineparser/commandline/pull/484#issuecomment-515363598)show the use case of help and it is passed)

                 var parseResult = parser.ParseArguments<Options_HelpText_Ordering_Verb1,    Options_HelpText_Ordering_Verb2>(
                new[] { "verb1", "--help" }) //call help
            .WithNotParsed(errors => { ; })
            .WithParsed(args => {; }); 
        
**Note**: This is the official way to test help: either by passing args '--help' or parser errors. It's not tested on parser success.
Also, you can use `ToNotEmptyLines()` method instead of splitting args using \r \n.

b3b00:
ok I didnot see the --help in test case. I correct it right now

b3b00:
@moh-hassan last build seems to have failed on upload failure. I can not restart and it does not seem to relate to my PR. Could you check it and maybe restart the build ?

moh-hassan:
I think the problem is temporary and it's related to CI  appveyor. 
The parameter `OptionComparison  in AutoBuild` isn't removed [here](https://github.com/commandlineparser/commandline/pull/484#discussion_r307653798). Can you update the commit [fix ](https://github.com/commandlineparser/commandline/pull/484/commits/cff9fd1b4c1cf9e6c8a0c86ebcf2dac5b40e9fd2) and remove the parameter?

b3b00:
@moh-hassan  : code seems ok, maybe you are looking at a stale version.

moh-hassan:
I looked at your last commit fix (cff9fd1) for file changed  [here ](https://github.com/commandlineparser/commandline/pull/484/files) and [here](https://github.com/b3b00/commandline/tree/feature/%23482).

Can you point me where can I look for the last update?

moh-hassan:
+1
Thanks Olivier Duhart for fixing.

b3b00:
@moh-hassan thank you for merging . Glad to have help a bit this project.

PR #492, opened 2019-08-06
Allow for Examples with just text or just sample

asherber:
I've had occasions where I want to output usage lines that have either text or a sample, but not both. This PR modified the `Example` constructors so that at least one of `helpText` and `sample` is required, rather than both.

This means that both of these are legitimate:
```csharp
new Example("Some text", null);
new Example(null, new Options() { Foo = "bar" });
```
An `Example` with text only will render the text without the trailing colon.

### DIFF
--- a/src/CommandLine/Text/Example.cs
+++ b/src/CommandLine/Text/Example.cs
@@ -23,9 +23,9 @@ namespace CommandLine.Text
         /// <param name="sample">A sample instance.</param>
         public Example(string helpText, IEnumerable<UnParserSettings> formatStyles, object sample)
         {
-            if (string.IsNullOrEmpty(helpText)) throw new ArgumentException("helpText can't be null or empty", "helpText");
             if (formatStyles == null) throw new ArgumentNullException("formatStyles");
-            if (sample == null) throw new ArgumentNullException("sample");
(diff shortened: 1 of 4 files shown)


moh-hassan:
The standard USAGE Example is composed using [FormatCommandLine ](https://github.com/commandlineparser/commandline/blob/master/src/CommandLine/UnParserExtensions.cs#L91)Extension method as given in your second use case  with optional  configuration of  UnParserSettings  like: `GroupSwitches, PreferShortName, UseEqualToken`, ...others and auto hide options when its value equal to the default value and may add quotation to the string options and other formatting features.

 This PR is a shortcut for passing string Expression for USAGE which may violate the rules of the UnParsing described above.
 It's recommended to use `UnParserExtensions` method to compose USAGE.

asherber:
What I'm looking for is a way to get these outputs from the default generated help:

```
USAGE:
This app is used to build frobbers.

// or 

USAGE:
  myapp --output myfile.txt
```

I don't think the current code allows me to do this.

I did change `RenderUsageTextAsLines()` so that it doesn't pass a null options argument to `FormatComandLine`, and all unit tests currently pass.

Issue #188, opened 2017-11-09
Immutable constructor argument order issues result in NullReferenceException


ohadschn:
For example:

```cs
    class FooOptions
    {
        public FooOptions(string option, Uri foo)
        {
            Foo = foo;
            Option = option;
        }

        [Option("foo")]
        public Uri Foo { get; }    

        [Option("option")]
        public string Option { get;  }
    }
```

Will result in:
```
nhandled Exception: System.NullReferenceException: Object reference not set to an instance of an object.
   at CommandLine.Infrastructure.ReflectionHelper.CreateDefaultImmutableInstance(Type type, Type[] constructorTypes)
   at CommandLine.Core.ReflectionExtensions.AutoDefault(Type type)
   at CommandLine.Core.InstanceChooser.<>c__DisplayClass1_0.<MatchVerb>b__1()
   at CommandLine.Core.InstanceBuilder.<>c__0`1.<Build>b__0_0(Func`1 f)
   at CSharpx.MaybeExtensions.MapValueOrDefault[T1,T2](Maybe`1 maybe, Func`2 func, T2 noneValue)
   at CommandLine.Core.InstanceBuilder.Build[T](Maybe`1 factory, Func`3 tokenizer, IEnumerable`1 arguments, StringComparer nameComparer, Boolean ignoreValueCase, CultureInfo parsingCulture, IEnumerable`1 nonFatalErrors)
   at CommandLine.Core.InstanceChooser.MatchVerb(Func`3 tokenizer, IEnumerable`1 verbs, IEnumerable`1 arguments, StringComparer nameComparer, CultureInfo parsingCulture, IEnumerable`1 nonFatalErrors)
   at CommandLine.Core.InstanceChooser.<>c__DisplayClass0_0.<Choose>b__0()
   at CommandLine.Core.InstanceChooser.Choose(Func`3 tokenizer, IEnumerable`1 types, IEnumerable`1 arguments, StringComparer nameComparer, CultureInfo parsingCulture, IEnumerable`1 nonFatalErrors)
   at CommandLine.Parser.ParseArguments(IEnumerable`1 args, Type[] types)
   at CommandLine.ParserExtensions.ParseArguments[T1,T2](Parser parser, IEnumerable`1 args)
   at CommandLineParserTest.Program.Main()
```

The ideal solution IMO would be to find the right ctor by param names (rather than order). But even if that's not implemented, a better and more meaningful exception would be very helpful in debugging such issues.

[...]

PR #495, opened 2019-08-07
 Fix NullReferenceException when creating a default immutable instance

0xced:


moh-hassan:
@0xced
Thanks for your effort.

Can your PR help in fixing the issue  #499?

0xced:
Ping.

moh-hassan:
Thanks for your work.
The Parser already throw `InvalidOperationException` not `NullReferenceException`  for default constructor.
What about if the  Exception message is  reworded and contain  something like:

>... the constructor should have parameters with the same names as Type properties and in the same order(inherited options are expected at the end of the constructor parameter list)..

Have a look to this [discussion](https://github.com/commandlineparser/commandline/issues/188)
This may help  in resolving immutable  types Ctor Exceptions.

0xced:
> The Parser already throw `InvalidOperationException` not `NullReferenceException` for default constructor.

Sorry, I should have properly explained the purpose of this pull request in the description; this is actually not always the case. When help is requested, the code takes another path and we get a `NullReferenceException` instead of `InvalidOperationException`. I have added `OptionClass_IsImmutable_HasNoCtor_HelpRequested` test to cover that specific issue.

I'll have a look at #188, but it looks orthogonal at first sight.

moh-hassan:
>When help is requested, the code takes another path

That is correct. But Parser isn't calling  the  `InstanceBuilder.Build` . It Fire the Help Screen.
`InvokeBuild` in Test  call ` InstanceBuilder.Build` .

The Parser still in help request fire `InvalidOperationException`. 

The attached file is a demo project to demonestrate help request with the Immutable class.
[ConsoleApp1.zip](https://github.com/commandlineparser/commandline/files/3896549/ConsoleApp1.zip)
run:

        consoleapp1 --help

**Edit:**
In case of using InvokeBuild, autohelp parameter should be false.

The next Test case Pass successfully.

```cs
 [Fact]
        public void Parse_Immutable_with_default_ctor_and_help__request_throw_InvalidOperationException()
        {

            Action act = () => InvokeBuild<ValueWithNoSetterOptions>(new string[] { "--help" }, false);
            act.Should().Throw<InvalidOperationException>();
        }
```

moh-hassan (approved):

Issue #486, opened 2019-07-31
How to contribute to Wiki Documentation


moh-hassan:
GitHub wikis are just Git repositories. Unfortunately contributors cannot fork the wikis through the GitHub UI, so while they can clone the wiki and make a change to the copy, they cannot submit a pull request.
A relatively easy workaround for this issue is to clone the wiki, do changes and send changes as Git Patches

Welcome to contributers to rich the wiki pages with examples

**How it works**

- Clone wiki documentation

      git clone https://github.com/commandlineparser/commandline.wiki.git

- create a new branch, e.g "custome-help"

             git checkout -b custome-help
   
- update/Add  wiki pages 

        git add .
       git commit -m "add custome-help.md"

- create Git Patch for your changes

      git format-patch master --stdout > custome-page.patch   

- Open a new Issue describing what changes in wiki , e.g
   Wiki: Add a new custom-page.md page
   Attach the patch file "custome-page.patch" as a zip file.

- Project maintainers will review the submitted patch file(s) of the wiki changes and merge it in the wiki Repo.

**Tips for wiki:**

- PR can contain wiki patche(s) for the new feature or fix with code example.
- You can provide code examples as an attached  zip file.
- You can use  [dotnetfiddle](https://dotnetfiddle.net) to show  a working example.
- Review the current open issues related to the documentation and submit a wiki page with examples.

Issue #353, opened 2018-10-23
Subverbs


frankcchen:
Does this library support subverbs? Based on my current research it doesn't. For example, let's assume git is the top level verb, And it has several layers of subverbs.

git/commit
git/push/folder  --> (file and folder are verbs as well)
git/push/file
git/pull/folder
git/pull/file

Is there anyway to write this kind of command line interface with multiple layers of subcommands/subverbs? Thanks

nemec:
It does not. There was a comment in one of the issues where someone mentioned they remove the first item from `args` each time a verb is hit and then run the parser over it again, but that seems kind of fragile.

Psuedocode Example:

```csharp
var args = new List<string>{"git", "push", "--help"};  // this comes from main
while(true)
{
    var result = Parser.Default.ParseArguments<GitVerb, PushVerb, PullVerb>(args);
    if(/* result is gitverb */)
    {
        args.Remove(0);
        continue;
    }
    RunProgram(result);
}
```

frankcchen:
@nemec  What if we treate the subverbs as values and after parsing we can use switch statement to determine what value(subcommand) it is and perform things accordingly. That might be a hack as well right?

```
[Verb("git"), HelpText = "git verb")]
class GitOptions
        {
            [Option("foo", Required = true, HelpText = "foo")]
            public string Name { get; set; }

            [Value(0)]
            public string SubVerb { get; set; }

            [Value(1)]
            public string NextVerb { get; set; }

        }
```

If we run` $git push file --foo "bar"`, opts.SubVerb is "push", opts.NextVerb is "file" and opts.foo is "bar".

ericnewton76:
nemecs comment seems legit.  I would use that functionality at least for the time being.

frankcchen, your solution will definitely get you in trouble because your sub verbs then wont get as good of handling the deeper you go.

Verbs that "consume" the args actually is an elegant solution for the problem until subverbs can be formally accounted for in the library. (Frankly this is probably how the library would handle sub verbs!... Lol)

simonech:
How would you handle the automatic generation of help for subverbs with this approach?

twaalewijn:
@simonech I recently needed to implement subverbs for a commandline tool I made so I put my solution on [GitHub](https://github.com/twaalewijn/commandlineparser-subverbextension).

The code to use it is a bit more complicated than the pseudo code provided by @nemec but it should handle the auto help/version well enough.

It also does not allow a user to pass, for example:
```
git git git git push --help
```
[...]

PR #552, opened 2019-12-08
Introduce option groups

### LINKED ISSUE #546: Could two or more options be marked as 'at least one is required'?
Hello,

This is not particularly an issue. I was wondering could I make two or more options to act like at least one of them is required, but if none is provided I will get a required error?

``` c#
[Verb("rename")]
public sealed class RenameOptions
{
    [Option("prefix", HelpText = "Prefix to append to file name", SetName = "rename", Required = true)]
    public string Prefix { get; set; }

    [Option("suffix", HelpText = "Suffix to append to file name", SetName = "rename", Required = true)]
    public string Suffix { get; set; }
}
```
This works if both options are included. Could I make them to work separately or together, but always one should be provided?

hadzhiyski:
I found a missing use case in the `commandline` library.
When one or more options has group set, at least one of these properties should have set value (they behave as required)

Example:

``` c#
[Option("option1", Required = true)]
public string Option1 { get; set; }

[Option("option2", Group="special")]
public string Option2 { get; set; }

[Option("option3", Group="special")]
public string Option3 { get; set; }
```

```
myapp --option1="test" --option2="asd" --option3="qwerty" // OK
myapp --option1="test" --option2="asd" // OK
myapp --option1="test" --option3="qwerty" // OK
myapp --option1="test" // Error: MissingGroupOptionError 
Error text: "At least one option in group 'special' must have value"
```

I hope this makes sense.

Closes #546


hadzhiyski:
Could you review this and approve this PR if you think it is worth introducing the feature? For me it is useful, because I need it for a project of mine - https://github.com/hadzhiyski/FileWarden/tree/develop

moh-hassan:
Thanks @hadzhiyski  for your work.

As I understand group (correct me):

- group is like Mutual Exclusive Options except in group at least one option in group is required(they need not to be defined as required).

- you can use more than one option in the same group (even all options ) at a time.

The following use case using `enum` may be equivalent for the group: 

```cs
class Options
{
[Option("option1", Required = true)]
public string Option1 { get; set; }

//use multi enum
[Option("special", Required = true)]
public IEnumerable<Special> SpecialOption { get; set; }
}

//stndardize option values using enum
enum Special 
{
asd,
qwerty
}
```
Mapping you test cases ,

This is valid command:
```
 myapp --option1 test --special asd  //ok only one special option
 myapp --option1 test --special qwerty  //ok only another one special option
 myapp --option1 test --special asd qwerty //ok both special options
 ```
 This is not valid:
```
 myapp --option1 test  //Error  no Special option is provided (required)
 ```
Can this approach help for your use case?

hadzhiyski:
Hello,

This is a useful solution. However it's not exactly what I am trying to achieve.
Imagine an application which appends prefix and/or suffix to a filename. Basically renames the file with prefix and/or suffix. Both options ("prefix" and "suffix") belong to the verb rename, so I am trying to achieve the following

```
myapp rename --source C:\test\file1.txt --prefix test --suffix _10122019 // OK. We have both options
myapp rename --source C:\test\file1.txt --prefix test // OK. We have prefix option only
myapp rename --source C:\test\file1.txt --suffix _10122019 // OK. We have suffix option only
```
But
```
myapp rename --source C:\test\file1.txt // Error: Neither prefix nor suffix is available
```

Mutually exclusive sets allow me use only one of the two options, but never together. 
Your solution might be useful if I want to pass multiple values to either to the prefix or suffix option.

How does this sound to you? I hope it makes sense.

moh-hassan:
I give it a quick run test.It seems it's useful and good.
Kindly, can you cover the following points:

- Provide unit test  that cover the different use-cases including exception and helpText.
- What about more than one group? It seems it support more than one (good).
- A wiki page to explain the value of the group with code samples?
- Is there a break change? It seems that there is a break change in some tests? 
- Can this feature help the concept of SubVerbs as explained [here ](https://github.com/commandlineparser/commandline/issues/353) (just for information, need not change in your PR)

hadzhiyski:
Brilliant!

I will try to do this next week, because I will be quite busy by the end of this week

hadzhiyski:
@moh-hassan How I could update the wiki? I cloned it, but when I try to push the changes, I am getting 'access denied' error. Could you add me to commandlineparser organisation or I can send you the changes here and you to update these for me later?

moh-hassan:
You can follow the instructions of [How to contribute to Wiki Documentation](https://github.com/commandlineparser/commandline/issues/486) and attach  the git patch file in PR.

hadzhiyski:
Hello,
Here are the wiki updates. [option-groups.zip](https://github.com/commandlineparser/commandline/files/3959761/option-groups.zip)

hadzhiyski:
- [X]  Provide unit test that cover the different use-cases including exception and helpText.
- [X] What about more than one group? It seems it support more than one (good).
>We can use a single group on a single option which basically acts like required. Multiple groups can be used on multiple options and they are separated from each other.
- [x] A wiki page to explain the value of the group with code samples?
- [X] Is there a break change? It seems that there is a break change in some tests?
> Group parameter is added to `OptionSpecification` and `OptionAttribute`
- [X] Can this feature help the concept of SubVerbs as explained here (just for information, need not change in your PR)
> In my opinion there is no connection between the two functionalities. Option groups are on 'option' level and sub verbs are on 'verb' level.

moh-hassan:
Thanks @hadzhiyski  for the update.
The wiki patch is applied successfully :) and it seems good.

There are some missing unit tests:

1.  Required
> Required rules are ignored if an option belongs to an option group. 

 Need a unit test

2.  Default Value
   a-Need a unit test 
    b-  It's failed and should success if no values is provided for the group
     > At least one option from group 'append' (prefix, suffix) is required.

  Group should take Default into account. 
  (if (option ==null) option = Default)

3.  What do you think for the priority of `SetName /Group`, which one can be ignored?
  [ ] ignore  SetName in favor of Group.
 [ ] ignore Group in favor of SetName.
 [ ] Raise Error: Only SetName or Group can be used. The two are not allowed.  (Preferred)

  4.  What is the effect of option with group if it's hide? (ignore all hiding options)
 
 Edit.

hadzhiyski:
> **Group should take Default into account.
> (if (option ==null) option = Default)**

@moh-hassan Could you give me more details about this? I can not understand what are you trying to say. Could you send a code sample or where this is located in source?

> **3. What do you think for the priority of SetName /Group, which one can be ignored?**
> Raise Error: Only SetName or Group can be used. The two are not allowed. 

I also think the best way will be to raise error.

moh-hassan:
@hadzhiyski 
>Group should take Default into account.
(if (option ==null) option = Default)

It means that when using Group for option and that Option has Default like:
```cs
[Option("option2", Group="special", Default="xyz")]
public string Option2 { get; set; }
```
Group should take the Default value "xyz" into account.
I suggest you a resolution for this issue using a Pseudo code rule (not included in your source and like c# code)
 ```
//Enforce this rule for the group
if (option.Value ==null) option.Value = Option.Default)
```
You are free to implement this rule.
I think you may be confused with Default because you are using this class name: [Simple_Options_With_OptionGroup_WithDefaultValue](https://github.com/commandlineparser/commandline/blob/4274c25b981e84d66fd1ed54c627e0b69d4d3a4f/tests/CommandLine.Tests/Fakes/Simple_Options_With_OptionGroup_WithDefaultValue.cs) which isn't related to Default :)

File: src/CommandLine/Core/OptionSpecification.cs [hunk 1]
```diff
@@ -13,16 +13,18 @@ sealed class OptionSpecification : Specification
         private readonly string longName;
         private readonly char separator;
         private readonly string setName;
+        private readonly Maybe<string> group;
```
moh-hassan:
```Suggestion
private readonly string group;
```
like `SetName`, No Null processing needed in group, just give it a name.
Need change type in Constructor and its references.

src/CommandLine/Core/OptionSpecification.cs
```diff
@@ -13,16 +13,18 @@ sealed class OptionSpecification : Specification
         private readonly string longName;
         private readonly char separator;
         private readonly string setName;
+        private readonly Maybe<string> group;
 
         public OptionSpecification(string shortName, string longName, bool required, string setName, Maybe<int> min, Maybe<int> max,
             char separator, Maybe<object> defaultValue, string helpText, string metaValue, IEnumerable<string> enumValues,
-            Type conversionType, TargetType targetType, bool hidden = false)
+            Type conversionType, TargetType targetType, Maybe<string> group, bool hidden = false)
```
moh-hassan: change `Maybe<string> group'   to=> `string group'

src/CommandLine/Core/OptionSpecification.cs
```diff
@@ -41,13 +43,14 @@ public static OptionSpecification FromAttribute(OptionAttribute attribute, Type
                 enumValues,
                 conversionType,
                 conversionType.ToTargetType(),
+                string.IsNullOrWhiteSpace(attribute.Group) ? Maybe.Nothing<string>() : Maybe.Just(attribute.Group),
```
moh-hassan:
```suggestion
 optionHelpText = "({0}: {1}) ".FormatInvariant(optionGroupWord, optionGroupSpecification.Group ??"" + optionHelpText);
```

src/CommandLine/Core/SpecificationPropertyRules.cs
```diff
@@ -16,12 +17,43 @@ static class SpecificationPropertyRules
...
 
+        private static Func<IEnumerable<SpecificationProperty>, IEnumerable<Error>> EnforceGroup()
+        {
+            return specProps =>
+            {
+                var optionsValues =
+                    from sp in specProps
+                    where sp.Specification.IsOption()
+                    let o = (OptionSpecification)sp.Specification
+                    where o.Group.IsJust()
```
moh-hassan:
```suggestion
  where o.Group.Length > 0
```

src/CommandLine/Core/SpecificationPropertyRules.cs
```diff
@@ -16,12 +17,43 @@ static class SpecificationPropertyRules
...
+                    select new
+                    {
+                        Option = o,
+                        Value = sp.Value
+                    };
+
+                var groups = from o in optionsValues
+                             group o by o.Option.Group.GetValueOrDefault(null) into g
```
moh-hassan:
```suggestion
 group o by o.Option.Group into g
```

src/CommandLine/Core/SpecificationPropertyRules.cs
```diff
@@ -81,6 +114,7 @@ where sp.Specification.Required
                             where sp.Value.IsNothing()
                             let o = (OptionSpecification)sp.Specification
                             where o.SetName.Length == 0
+                            where o.Group.IsNothing()
```
moh-hassan:
```suggestion
 where o.Group.Length == 0
```

src/CommandLine/Core/OptionSpecification.cs
moh-hassan:
```suggestion
 attribute.Group??string.Empty,
 ```

File: src/CommandLine/Core/OptionSpecification.cs [hunk 7]
```diff
@@ -41,13 +43,14 @@ public static OptionSpecification FromAttribute(OptionAttribute attribute, Type
...
                 conversionType.ToTargetType(),
+                string.IsNullOrWhiteSpace(attribute.Group) ? Maybe.Nothing<string>() : Maybe.Just(attribute.Group),
                 attribute.Hidden);
         }
 
         public static OptionSpecification NewSwitch(string shortName, string longName, bool required, string helpText, string metaValue, bool hidden = false)
         {
             return new OptionSpecification(shortName, longName, required, string.Empty, Maybe.Nothing<int>(), Maybe.Nothing<int>(),
-                '\0', Maybe.Nothing<object>(), helpText, metaValue, Enumerable.Empty<string>(), typeof(bool), TargetType.Switch, hidden);
+                '\0', Maybe.Nothing<object>(), helpText, metaValue, Enumerable.Empty<string>(), typeof(bool), TargetType.Switch, Maybe.Nothing<string>(), hidden);
```
moh-hassan: replace the Group parameter: `Maybe.Nothing<string>()`
to
`string.Empty`

moh-hassan:
@hadzhiyski 
Kindly, Stop changing your PR because I start reviewing yours.
I will ,a little time, send you the comments of review with the suggested changes.

src/CommandLine/Text/HelpText.cs
```diff
@@ -865,8 +871,18 @@ private HelpText AddPreOptionsLine(string value, int maximumLength)
             return this;
         }
 
-        private HelpText AddOption(string requiredWord, int maxLength, Specification specification, int widthOfHelpText)
+        private HelpText AddOption(string requiredWord, string optionGroupWord, int maxLength, Specification specification, int widthOfHelpText)
         {
+            OptionSpecification GetOptionGroupSpecification()
+            {
+                if (specification.Tag == SpecificationType.Option && specification is OptionSpecification optionSpecification && optionSpecification.Group.IsJust())
```
moh-hassan:
```suggestion
  if (specification.Tag == SpecificationType.Option &&
                    specification is OptionSpecification optionSpecification &&
                    !string.IsNullOrEmpty(optionSpecification.Group))
```

moh-hassan (commented):
I suggest change the type of Group from  Maybe<string> to string, for the following reasons:
- Group is definded in OptionAttribute Class as [string](https://github.com/commandlineparser/commandline/blob/4274c25b981e84d66fd1ed54c627e0b69d4d3a4f/src/CommandLine/OptionAttribute.cs#L19) and [here](https://github.com/commandlineparser/commandline/blob/4274c25b981e84d66fd1ed54c627e0b69d4d3a4f/src/CommandLine/OptionAttribute.cs#L112)
- It's like `string SetName` which is initialized to string.Empty
- There is no null processing /or Mapping in the Group.It's just only a name.  
- You need not to do such expressions: 

          string.IsNullOrWhiteSpace(attribute.Group) ? Maybe.Nothing<string>() : Maybe.Just(attribute.Group),
 
 
I did these modification in my side and test pass except 3 tests in the multi group(not yet reviewed).

src/CommandLine/Text/HelpText.cs
```diff
@@ -889,13 +905,20 @@ private HelpText AddOption(string requiredWord, int maxLength, Specification spe
...
-            if (specification.Required)
+            var optionGroupSpecification = GetOptionGroupSpecification();
+
+            if (specification.Required && optionGroupSpecification == null)
                 optionHelpText = "{0} ".FormatInvariant(requiredWord) + optionHelpText;
-          
+
+            if (optionGroupSpecification != null)
+            {
+                optionHelpText = "({0}: {1}) ".FormatInvariant(optionGroupWord, optionGroupSpecification.Group.GetValueOrDefault(null)) + optionHelpText;
```
moh-hassan:
```suggestion
  optionHelpText = "({0}: {1}) ".FormatInvariant(optionGroupWord, optionGroupSpecification.Group)  + optionHelpText;
```

moh-hassan (commented):
All Test passed including the 3 multi group help test.

If you want, I can update your PR with a commit: `"Modify type of Group to string instead of Mayb<string>"`

hadzhiyski:
@moh-hassan Thanks for the review.
If you already have the commit, please do it. Otherwise I will make the changes.

moh-hassan:
PR is updated by the [commit: Modify type of Group to string instead of `Maybe<string>' ](https://github.com/commandlineparser/commandline/pull/552/commits/ff61b69a19a870b3ec1e905eee4c2e319ae00d0a)

hadzhiyski:
Brilliant. Thank you so much!

moh-hassan (approved):

hadzhiyski:
@moh-hassan Thank you for the great work! 
In which version I could expect this feature to be included?

sm-g:
@moh-hassan cli is confusing now

for such options:
```
internal class Options
{
	[Option('u', "url", SetName = "url", Group = "input", Default = "http://localhost:5002/swagger/current/swagger.json")]
	public string SwaggerUrl { get; set; }

	[Option('j', "json", SetName = "json", Group = "input")]
	public string SwaggerJson { get; set; }

	[Option('f', "file", SetName = "file", Group = "input")]
	public string SwaggerFilePath { get; set; }
}
```
when running exe without any args we have output:
```
 At least one option from group 'input' (u, url, j, json, f, file) is required.

  -u, --url       (Group: input) (Default: http://localhost:5002/swagger/current/swagger.json)

  -j, --json      (Group: input)

  -f, --file      (Group: input)
```

So, default value for url will be set only when exe invoked with `exe -j "value"` or `exe -f "value"`. Which is not what I expect as user of program when see help message.

Also new Group feature does not prevent setting default value for "mutually exclusive" properties, it that by design?

moh-hassan:
@sm-g 
Both SetName and group are not allowed for the option in the same time.  Use only SetName OR Group.

Removing SetName
```cs
internal class Options
{
	[Option('u', "url",  Group = "input", Default = "http://localhost:5002/swagger/current/swagger.json")]
	public string SwaggerUrl { get; set; }

	[Option('j', "json",  Group = "input")]
	public string SwaggerJson { get; set; }

	[Option('f', "file", Group = "input")]
	public string SwaggerFilePath { get; set; }
}
```
a new PR may Support Default Value for Group.

cc @hadzhiyski 
It's nice if Group support Default value and firing error message when both **SetName** and **Group** are used in the same time.

hadzhiyski:
@moh-hassan, @sm-g I will submit a new PR for this.

moh-hassan:
Thanks @hadzhiyski for the quick response and support.

hadzhiyski:
The new PR is #575

PR #560, opened 2020-01-08
Tests cleanup

gsscoder:
Just some few order in test code.

### DIFF
--- a/tests/CommandLine.Tests/Unit/UnParserExtensionsTests.cs
+++ b/tests/CommandLine.Tests/Unit/UnParserExtensionsTests.cs
@@ -3,6 +3,7 @@
 using System;
 using System.Collections.Generic;
 using System.Linq;
+using CommandLine.Tests.Fakes;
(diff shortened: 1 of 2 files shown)


tests/CommandLine.Tests/Unit/UnParserExtensionsTests.cs
```diff
@@ -3,11 +3,9 @@
 using System;
 using System.Collections.Generic;
 using System.Linq;
-using System.Text.RegularExpressions;
-using CommandLine.Tests.Fakes;
 using Xunit;
 using FluentAssertions;
-using Microsoft.FSharp.Core;
+using CommandLine.Tests.Fakes;
```
moh-hassan: //Microsoft.FSharp.Core is needed for fsharp Target_Build. This is the reason for appveyor fail
```suggestion
#if !SKIP_FSHARP
using Microsoft.FSharp.Core;
#endif
```

PR #575, opened 2020-02-02
Fix option groups

hadzhiyski:
- Option groups take into account default value
- Option groups can not be used with mutually exclusive set together
- Multiple group errors are shown together

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,12 +1,12 @@
 #version should be only changed with RELEASE eminent, see RELEASE.md
 
-version: 2.7.83-beta-{build}
+version: 2.7.84-beta-{build}
 
 image: Visual Studio 2019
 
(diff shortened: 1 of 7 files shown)


moh-hassan:
@hadzhiyski ,
I tried the attached project after merging
Default is working fine.
When group and setname are included in the option,  I get an Exception "Operation not allowed". The details of error is included in the sample.
Can you check this case.

[issues.zip](https://github.com/commandlineparser/commandline/files/4145058/issues.zip)

moh-hassan:
I had [fixed ](https://github.com/commandlineparser/commandline/commit/5bb3235e42c2a81b7d300da9ec880e1941c3f924)the Exception issue. No action is needed.
Thanks @hadzhiyski  for the PR.

Issue #70, opened 2017-11-04
Parser.ParseArguments(factory, args) requires parameterless constructor


ericnewton76:
<a href="https://github.com/Virtlink"><img src="https://avatars0.githubusercontent.com/u/647530?v=4" align="left" width="96" height="96" hspace="10"></img></a> **Issue by [Virtlink](https://github.com/Virtlink)**
_Sunday Apr 24, 2016 at 11:12 GMT_
_Originally opened as https://github.com/gsscoder/commandline/issues/310_

----

The `Parser.ParseArguments()` overloads with a factory function has this signature:

```
public ParserResult<T> ParseArguments<T>(Func<T> factory, IEnumerable<string> args)
    where T : new()
```

If I provide my own factory function, why does `T` need to have a _public_ parameterless constructor (`new()`)?

```
var result = parser.ParseArguments<Options>(() => new Options(true), args);
```

And on the other hand, why doesn't the non-factory overload of `ParseArguments` _not_ require a public parameterless constructor? This will cause a runtime-exception when `T` doesn't have a public parameterless constructor.

```
var result = parser.ParseArguments<Options>(args);
```

ericnewton76:
<a href="https://github.com/nemec"><img src="https://avatars3.githubusercontent.com/u/136378?v=4" align="left" width="48" height="48" hspace="10"></img></a> **Comment by [nemec](https://github.com/nemec)**
_Monday Apr 25, 2016 at 01:38 GMT_

----

I can't speak to the reasoning behind the decision, since I didn't write the code, but they both look like oversights to me. If you want to submit a pull request, I'll merge it.

ericnewton76:
<a href="https://github.com/Virtlink"><img src="https://avatars0.githubusercontent.com/u/647530?v=4" align="left" width="48" height="48" hspace="10"></img></a> **Comment by [Virtlink](https://github.com/Virtlink)**
_Monday Apr 25, 2016 at 10:56 GMT_

----

I can fix the first by removing the `where T : new()`, but I'm not sure about the semantics of the second as adding `where T : new()` causes tests to fail.

ericnewton76:
<a href="https://github.com/JeanSebTr"><img src="https://avatars1.githubusercontent.com/u/867435?v=4" align="left" width="48" height="48" hspace="10"></img></a> **Comment by [JeanSebTr](https://github.com/JeanSebTr)**
_Wednesday Apr 27, 2016 at 18:43 GMT_

----

[...]

PR #590, opened 2020-03-07
Remove constraint on T for ParseArguments with factory

pergardebrink:
Fix for https://github.com/commandlineparser/commandline/issues/70 and https://github.com/commandlineparser/commandline/issues/591

This makes it possible to have mutable types without empty constructor that only does explicit implementation of interfaces, like this (the IsMutable() check would fail previously in this particular case):
```
public interface IFoo {
    [Option("foobar")]
    string FooBar { get; set; }
}
public interface IBar {
    [Option("barfoo")]
    string BarFoo { get; set; }
}

public class FooBarOptions : IFoo, IBar {
    public FooBarOptions(string nonEmptyConstructor) { }
    IBar.BarFoo { get; set; }
    IFoo.FooBar { get; set; }
}

var result = Parser.Default.ParseArguments<FooBarOptions>(() => new FooBarOptions("foo"), args)
```

### DIFF
--- a/src/CommandLine/Core/InstanceBuilder.cs
+++ b/src/CommandLine/Core/InstanceBuilder.cs
@@ -39,7 +39,7 @@ namespace CommandLine.Core
 
             Func<T> makeDefault = () =>
                 typeof(T).IsMutable()
-                    ? factory.MapValueOrDefault(f => f(), Activator.CreateInstance<T>())
+                    ? factory.MapValueOrDefault(f => f(), () => Activator.CreateInstance<T>())
                     : ReflectionHelper.CreateDefaultImmutableInstance<T>(
                         (from p in specProps select p.Specification.ConversionType).ToArray());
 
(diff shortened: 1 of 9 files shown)


moh-hassan:
Thanks @pergardebrink for your effort.
Nice that you fixed this missing issue.

moh-hassan (approved):

moh-hassan:
Just for correcting typo for the example in your post :
```cs
public class FooBarOptions : IFoo, IBar 
{
    public FooBarOptions(string nonEmptyConstructor) { }     
   string  IBar.BarFoo {get;set;}
   string  IFoo.FooBar { get; set; }   
}
```

pergardebrink:
> Thanks @pergardebrink for your effort.
> Nice that you fixed this missing issue.

No problem, and thanks for accepting the PR! 

I have another use case than my example that relied on having this working, so I was hoping to avoid forking the project just for that reason! ;)

PR #597, opened 2020-03-12
fix issue #593

### LINKED ISSUE #593: Unit tests fail if terminal window is not exactly 80 characters wide
https://github.com/commandlineparser/commandline/pull/589 adds a check for `sut.Settings.MaximumDisplayWidth.Should().Be(80)`, but that will fail if you run `dotnet test` in a Linux terminal window that's a different size.

### To Reproduce:

1. Clone the repo
1. Open a Linux terminal window and don't resize it, so it's 80x24 (the usual default)
1. Run `dotnet test`. It works.
1. Resize the terminal window so it's one character wider
1. Run `dotnet test`. It fails with "Expected sut.Settings.MaximumDisplayWidth to be 80, but found 81."

Note that on my dev laptop, I habitually run terminal windows at maximum screen width, so for me the exception was "Expected sut.Settings.MaximumDisplayWidth to be 80, but found 272."

### Solution:

Don't assume any particular value for MaximumDisplayWidth; just check that it's greater than 1.

moh-hassan:
Fix #593

### DIFF
--- a/tests/CommandLine.Tests/Unit/ParserTests.cs
+++ b/tests/CommandLine.Tests/Unit/ParserTests.cs
@@ -902,7 +902,7 @@ namespace CommandLine.Tests.Unit
             var sut = new Parser(config => config.HelpWriter = null);
             sut.ParseArguments<Simple_Options>(new[] {"--dummy"});
             //Assert
-            sut.Settings.MaximumDisplayWidth.Should().Be(80);
+            sut.Settings.MaximumDisplayWidth.Should().BeGreaterThan(1);
         }
     }
 }


moh-hassan:
@ericnewton76 
@gsscoder 
@nemec 

Hi all,
Nuget key is expired since more than month and no new packages can be published and fire  exception error and cause failure of appveyor.
version 2.8 can't be published and it was planed to be published in 17/3/2020.
I have no grant to update nuget key.
Kindly, can you   upgrade the nuget key and let me know to publish the package.
Thanks.

PR #594, opened 2020-03-12
Add multi-instance option support

rmunn:
https://github.com/commandlineparser/commandline/issues/357 is a request for supporting multiple occurrences of the same parameter and collecting those into an enumerable, e.g. "myprog -i file1 -i file2 -i file3" would parse as having the `-i` parameter with three values, file1, file2 and file3. @tydunkel created an implementation of this feature, but never (AFAICT) created a PR for it. This PR is @tydunkel's code, rebased onto the current `develop` branch.

Note that when I run this, I get some spurious unit test failures and some real failures. The spurious failures I'm getting are caused by https://github.com/commandlineparser/commandline/issues/403, but the real failures I'm getting all have the same pattern: the "ERROR(S)" section of the output is repeated twice. E.g., [this test](https://github.com/commandlineparser/commandline/blob/017a8a21c79b223fbcd541187c7dd546c18ebdda/tests/CommandLine.Tests/Unit/ParserTests.cs#L523) expects the parser to output:

```plaintext
testhost 15.3.0
Copyright (C) 2020 author
ERROR(S):
No verb selected.
add        Add file contents to the index.
commit     Record changes to the repository.
clone      Clone a repository into a new directory.
help       Display more information on a specific command.
version    Display version information.
```

and instead it's outputting:

```plaintext
testhost 15.3.0
Copyright (C) 2020 author
ERROR(S):
No verb selected.
ERROR(S):
No verb selected.
add        Add file contents to the index.
commit     Record changes to the repository.
clone      Clone a repository into a new directory.
help       Display more information on a specific command.
version    Display version information.
```

I don't have time right now to track down why the error output is being duplicated. But all the unit test failures I could see look like they're the same thing, so once that single bug is located and fixed, the tests should all pass.

Hopefully this will be helpful to someone to be able to move forward with implementing #357 (and #522 which is a dupe of #357).

PR #607, opened 2020-03-27
Fix #601 - better getopt compatibility for -- and --help handling

### LINKED ISSUE #601: Corner cases where getopt behavior is not mimiced: -- or --help as string values
The goal of this library is to [mimic the behavior of getopt](https://github.com/commandlineparser/commandline/issues/498#issuecomment-521757260), but there are a few corner cases where this library behaves differently than getopt would: in the handling of `--` or `--help` when they are the value of a string parameter.

## How getopt behaves

First, an illustration of how getopt works with the particular corner case I'm demonstrating. Let's look at the standard `gzip` and `gunzip` tools found with any Linux distribution. They take many options, but one of them is `--suffix` (or `-S` for short); this lets you specify a different suffix than the standard `.gz` for the compressed file. E.g. if you have a README.md file in the current directory, then `gzip -S .compressed README.md` will create a README.md.compressed file instead of README.md.gz.

Now, what do you think will happen if I run this command?

```plaintext
gzip -S -- README.md
```

The correct answer is that it will create a compressed file named `README.md--` in the current directory. Because the string `--` was specified immediately after an option that takes a string value, it was processed as the value for that option (the `--suffix` option), and so gzip created a file with a `--` suffix instead of `.gz`. Now look at these three examples:

```plaintext
1. gzip -- --help
2. gzip -S -- --help
3. gzip -S -- -- --help
```

What do you think these will do? Answer:

1. This will compress a file named `--help` in the current directory, and create a file named `--help.gz`.
2. This will print the help text, and do nothing else.
3. This will compress a file named `--help` in the current directory, and create a file named `--help--`.

Why did `gzip -S -- --help` print the help text? Because `--` was the value for the `-S` option, and so it was *not* treated as the "stop processing options now" marker. Then after the `-S` option was fully processed, the only remaining options were `--help`. Since `--help` was encountered, gzip displayed the help screen and did nothing else.

With the `gzip -S -- -- --help` line, OTOH, the first `--` became the value for the `-S` option. Then the second `--` was processed as an option, and had the "stop processing options now" meaning. So the `--help` text was treated as a value, and so it looked for a file named `--help` to compress. And since I specified that the compressed suffix should be `--`, the compressed file was named `--help--`.

## What CommandLine does

The current way CommandLine works is to [call a preprocessor function](https://github.com/commandlineparser/commandline/blob/3e43750c5e0c3a433c61d8b7df0bffb8cb9d76ec/src/CommandLine/Core/Tokenizer.cs#L46) to look for any `--` options and, if found, mark anything found after them as a value. But this would mean that in the `gzip -S -- --help` example, where the correct getopt-mimicing behavior would be to print the help text, CommandLine will instead return an error saying that `-S` needed a value and didn't get one.

This corner case actually shows a fundamental difference between the behavior of CommandLine and the behavior of getopt. CommandLine uses a tokenizer to parse the command-line arguments and decide, based on the presence of `-` or `--` at the front, to treat them as Name tokens or Value tokens. But if you [read the `getopt` source code](https://github.com/gcc-mirror/gcc/blob/master/libiberty/getopt.c) and figure out what it's actually doing, it's parsing one argument at a time, deciding whether that argument needs a value, and then if a value is needed, it swallows the next argument without further processing. Which is why you can pass `--` as the suffix in gzip, and it will happily accept that.

## What CommandLine should do

The tokenizer, instead of processing all the arguments at once and deciding whether they're names or values, should process each argument one at a time. Then the decision tree should look like:

- Is this option exactly `--` and EnableDashDash is true? Then stop processing; the rest of the arguments are all values.
- Is this option exactly `--` and EnableDashDash is false? Then it is the value `--`; continue processing the next argument.
- Does this option start with `--` and contain an equals sign? Then split it into two tokens, the part before the `=` is the name, and the part after the equals is the value. (Split at the first equals sign; any equals signs after that point would become part of the value).
- Does this option start with `--` and not contain an equals sign? Then we look at the list of option longnames that the tokenizer was given:
    - Name matches a boolean option: this is a name token. Resume tokenizing with the next argument (it is NOT swallowed).
    - **NEW FEATURE**: Name matches an int option and the option attribute has `AllowMultiple=true`: this is a name token. Resume tokenizing with the next argument (it is NOT swallowed). (This allows for things like `-v` or `--verbose` to be passed multiple times, like `-vvv`, which the parser will turn into Verbose=3 in the final options instance.)
    - Name matches an option that's neither of the two cases above (boolean or int with AllowMultiple): this is a name token, and the next argument is a value token *no matter what it is*. "Swallow" the next argument, and resume tokenizing with the argument after next.
- Does the option start with `-` and contain only letters that match shortnames? Split it into multiple shortnames. (I.e., `-lR` would become `Name("l"), Name("R")` if there are both `-l` and `-R` options).
- Does the option start with `-` and its first letter matches a shortname, but the rest does not? Split it into first letter & rest, and that's two tokens: Name(first letter) and Value(rest).
- Does the option start with `-` and have only one letter? Then it's a shortname, and we look at the type of the option with that shortname:
    - As above, if boolean, then don't swallow the next argument.
    - NEW FEATURE: As above, if int with AllowMultiple, then don't swallow the next argument.
    - As above, if other type, then swallow the next argument (WHATEVER it is) and treat it as a value.

## Conclusion

Unfortunately, if the goal of getopt compatibility is to be achieved, a big rewrite of the guts of CommandLine's tokenizer and parser will be needed, so this is a big job. But if we want to mimic the behavior of getopt, then that's what will be needed. And the behavior I described above is how getopt works.

Also unfortunately, this is probably going to be a breaking change, so it might end up requiring a 3.0 version number. Because some people might be very surprised when `--stringoption --booloption` ends up being parsed with `--booloption` as the string value of `--stringoption`; they would probably have come to expect that to produce a `MissingValueOptionError` for `--stringoption`. But surprise or not, the correct way to handle that is for `--booloption` to be the string value of `--stringoption` in that example.

rmunn:
This changes the tokenizer to use the same logic as glibc's getopt, where the handling of one token will influence the handling of the next token. In particular, *anything* that follows an option expecting a value, even if it is something specially handled, will become the value of that option. So if `-s` is an option expecting a string value, then `myprog -s --help` will NOT display the help screen, but will instead give `-s` the value `"--help"`, with no `--` needed.

To prove that this is how getopt behaves (and is expected to behave), run the following on any Linux machine: `gzip -S --help README.md`. The `-S` option lets you pick a suffix different from `.gz`, so if there's a README.md file in the current directory, this will create a gziped `README.md--help` file. This PR adjusts the behavior of CommandLineParser to mimic getopt's behavior exactly in this regard.

### DIFF
--- a/src/CommandLine/OptionAttribute.cs
+++ b/src/CommandLine/OptionAttribute.cs
@@ -17,6 +17,7 @@ namespace CommandLine
         private string setName;
         private char separator;
         private string group=string.Empty;
+        private bool flagCounter;
 
         private OptionAttribute(string shortName, string longName) : base()
         {
@@ -27,6 +28,7 @@ namespace CommandLine
             this.longName = longName;
             setName = string.Empty;
             separator = '\0';
+            flagCounter = false;
         }
 
         /// <summary>
@@ -113,6 +115,15 @@ namespace CommandLine
         {
             get { return group; }
             set { group = value; }
+        }
+
+        /// <summary>
+        /// When applied to an int property, turns that property into a count of how many times a boolean flag was applied (e.g., -vvv would become 3)
+        /// </summary>
+        public bool FlagCounter
+        {
+            get { return flagCounter; }
+            set { flagCounter = value; }
         }
     }
 }
--- a/src/CommandLine/Core/InstanceBuilder.cs
+++ b/src/CommandLine/Core/InstanceBuilder.cs
@@ -89,14 +89,14 @@ namespace CommandLine.Core
                     OptionMapper.MapValues(
                         (from pt in specProps where pt.Specification.IsOption() select pt),
                         optionsPartition,
-                        (vals, type, isScalar) => TypeConverter.ChangeType(vals, type, isScalar, parsingCulture, ignoreValueCase),
+                        (vals, type, isScalar, isFlag) => TypeConverter.ChangeType(vals, type, isScalar, isFlag, parsingCulture, ignoreValueCase),
                         nameComparer);
 
                 var valueSpecPropsResult =
                     ValueMapper.MapValues(
                         (from pt in specProps where pt.Specification.IsValue() orderby ((ValueSpecification)pt.Specification).Index select pt),
                         valuesPartition,
-                        (vals, type, isScalar) => TypeConverter.ChangeType(vals, type, isScalar, parsingCulture, ignoreValueCase));
+                        (vals, type, isScalar) => TypeConverter.ChangeType(vals, type, isScalar, false, parsingCulture, ignoreValueCase));
 
                 var missingValueErrors = from token in errorsPartition
                                          select
--- a/src/CommandLine/Core/InstanceBuilder.cs
+++ b/src/CommandLine/Core/InstanceBuilder.cs
@@ -24,6 +24,31 @@ namespace CommandLine.Core
             bool autoVersion,
             IEnumerable<ErrorType> nonFatalErrors)
         {
+            return Build(
+                factory,
+                tokenizer,
+                arguments,
+                nameComparer,
+                ignoreValueCase,
+                parsingCulture,
+                autoHelp,
+                autoVersion,
+                false,
+                nonFatalErrors);
+        }
+
+        public static ParserResult<T> Build<T>(
+            Maybe<Func<T>> factory,
+            Func<IEnumerable<string>, IEnumerable<OptionSpecification>, Result<IEnumerable<Token>, Error>> tokenizer,
+            IEnumerable<string> arguments,
+            StringComparer nameComparer,
+            bool ignoreValueCase,
+            CultureInfo parsingCulture,
+            bool autoHelp,
+            bool autoVersion,
+            bool allowMultiInstance,
+            IEnumerable<ErrorType> nonFatalErrors)
+        {
             var typeInfo = factory.MapValueOrDefault(f => f().GetType(), typeof(T));
 
             var specProps = typeInfo.GetSpecifications(pi => SpecificationProperty.Create(
@@ -70,7 +95,7 @@ namespace CommandLine.Core
                 var valueSpecPropsResult =
                     ValueMapper.MapValues(
                         (from pt in specProps where pt.Specification.IsValue() orderby ((ValueSpecification)pt.Specification).Index select pt),
-                        valuesPartition,    
+                        valuesPartition,
                         (vals, type, isScalar) => TypeConverter.ChangeType(vals, type, isScalar, parsingCulture, ignoreValueCase));
 
                 var missingValueErrors = from token in errorsPartition
@@ -86,7 +111,7 @@ namespace CommandLine.Core
 
                 //build the instance, determining if the type is mutable or not.
                 T instance;
-                if(typeInfo.IsMutable() == true)
+                if (typeInfo.IsMutable() == true)
                 {
                     instance = BuildMutable(factory, specPropsWithValue, setPropertyErrors);
                 }
@@ -95,7 +120,7 @@ namespace CommandLine.Core
                     instance = BuildImmutable(typeInfo, factory, specProps, specPropsWithValue, setPropertyErrors);
                 }
 
-                var validationErrors = specPropsWithValue.Validate(SpecificationPropertyRules.Lookup(tokens));
+                var validationErrors = specPropsWithValue.Validate(SpecificationPropertyRules.Lookup(tokens, allowMultiInstance));
 
                 var allErrors =
                     tokenizerResult.SuccessMessages()
--- a/src/CommandLine/Core/InstanceChooser.cs
+++ b/src/CommandLine/Core/InstanceChooser.cs
@@ -23,6 +23,31 @@ namespace CommandLine.Core
             bool autoVersion,
             IEnumerable<ErrorType> nonFatalErrors)
         {
+            return Choose(
+                tokenizer,
+                types,
+                arguments,
+                nameComparer,
+                ignoreValueCase,
+                parsingCulture,
+                autoHelp,
+                autoVersion,
+                false,
+                nonFatalErrors);
+        }
+
+        public static ParserResult<object> Choose(
+            Func<IEnumerable<string>, IEnumerable<OptionSpecification>, Result<IEnumerable<Token>, Error>> tokenizer,
+            IEnumerable<Type> types,
+            IEnumerable<string> arguments,
+            StringComparer nameComparer,
+            bool ignoreValueCase,
+            CultureInfo parsingCulture,
+            bool autoHelp,
+            bool autoVersion,
+            bool allowMultiInstance,
+            IEnumerable<ErrorType> nonFatalErrors)
+        {
             var verbs = Verb.SelectFromTypes(types);
             var defaultVerbs = verbs.Where(t => t.Item1.IsDefault);
             
@@ -46,7 +71,7 @@ namespace CommandLine.Core
                             arguments.Skip(1).FirstOrDefault() ?? string.Empty, nameComparer))
                     : (autoVersion && preprocCompare("version"))
                         ? MakeNotParsed(types, new VersionRequestedError())
-                        : MatchVerb(tokenizer, verbs, defaultVerb, arguments, nameComparer, ignoreValueCase, parsingCulture, autoHelp, autoVersion, nonFatalErrors);
+                        : MatchVerb(tokenizer, verbs, defaultVerb, arguments, nameComparer, ignoreValueCase, parsingCulture, autoHelp, autoVersion, allowMultiInstance, nonFatalErrors);
             };
 
             return arguments.Any()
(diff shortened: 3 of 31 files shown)


rmunn:
Note that I wrote this PR on top of #594, because creating lists of input files with `-i file1 -i file2 -i file3` is how getopt is supposed to work, and the current behavior of `-i file1 file2 file3` is *not* getopt-compatible. I would eventually like to see `-i file1 file2 file3` become an option that has to be explicitly enabled, and #594 become how CommandLineParser works by default.

This means that if you merge this PR as-is, #594 will automatically be merged as well.

moh-hassan:
@rmunn 
Thanks to your work. 
>This means that if you merge this PR as-is, #594 will automatically be merged as well.

and also #610 will automatically be merged as well.

src/CommandLine/OptionAttribute.cs
```diff
@@ -114,5 +116,14 @@ public string Group
             get { return group; }
             set { group = value; }
         }
+
+        /// <summary>
+        /// When applied to an int property, turns that property into a count of how many times a boolean flag was applied (e.g., -vvv would become 3)
+        /// </summary>
+        public bool FlagCounter
```
moh-hassan, line 123: Provide Test case for testing FlagCounter.
It seems it's not working and raise error when trying to use a commandline: '-vvv'  or "-v -v -v" with int property
```cs
 class Options
        {
            [Option('v',FlagCounter =true)]
            public int Visible { get; set; } //expect to be 3
        }
```

moh-hassan:
It is better to minimize the change in the public API to avoid break change ( except it is necessary), can we remove `FlagCounter` and it is considered as enabled when multiInstance is true, especially it's special case for multiInstance.

rmunn:
`FlagCounter` is a necessary addition to enable the `-vv` scenario, otherwise since it's defined as an `int` property it would be assumed to expect a value, so `-vv` would try to parse the second `v` as an int and would fail. It's necessary to add a flag so that options that are *expected* to be flag-like but passed multiple times (e.g., `-v` for verbose) can be distinguished from options that are *expected* to be passed only once but take a value (e.g. `-l 5` for "limit 5" in Git logs). Setting `FlagCounter` to be considered enabled when multiInstance is true would cause the `-l 5` scenario to fail, as there would then be an ambiguity between "assign 5 to the `-l` option" and "assign 1 to the `-l` option since it's only specified once, and then assign 5 to the next Value in line".

I'll add tests for FlagCounter. The tests I just pushed are to test the scenarios that came up in discussion here: https://github.com/commandlineparser/commandline/pull/610#issuecomment-636692036

rmunn:
I'm going to remove the FlagCounter feature from this PR: it's proving more difficult to implement than I had anticipated, and is going to require some fundamental reshuffling of how flag options are handled (a FlagCounter option needs to be treated more like a sequence than a flag throughout most of the code, up until the point where values are assigned). The required code changes to implement FlagCounter are making this PR drift out of its scope, so I'm going to pull them out of this PR and submit a separate PR for the FlagCounter feature once I get it working.

src/CommandLine/OptionAttribute.cs
rmunn: I ended up removing the FlagCounter feature. The work I've done on it so far, including a test case, is in the https://github.com/rmunn/commandline/tree/wip/flagCounter branch of my repo. As suggested by the `wip` in the name, it's a work-in-progress that is not yet ready for a PR or merge, and I may squash or force-push that branch at any time.

But with that feature removed from this PR, this particular code review is resolved.

rmunn:
Update: turns out I was almost there with my FlagCounter implementation yesterday, and I just needed a one-line change to get it all working. So I'm reverting my reversion of the feature, backing commit b8b9a97 out of this PR, and replacing it with commit 1e791b5 that implements FlagCounter properly and has a unit test to verify that it's working.

@moh-hassan - With this commit, I believe I've addressed all the things you mentioned in your code review and it's ready to merge if you're satisfied. With the new FlagCounter implementation, though, you might want to do another code review. If so, I'll be happy to address any further changes you might want me to make.

moh-hassan:
@rmunn 
I try to finalize both #594 and #610 then merging to develop branch  because your PR  depends on them.

I'm still in the code review for #607. 
Kindly can you reply on my [comment](https://github.com/commandlineparser/commandline/issues/601#issuecomment-635043612) especially note 2, 3 and 4 because there is a major change in `Tokenizer` and may cause a break change and moving to V3.

rmunn:
@moh-hassan - I had missed that comment. I'm reading it now and will respond to it over there.

This PR includes #594 so if you merge this one, #594 will automatically be merged as well. This one also contains a *rewrite* of #610 (with credit for the relevant commit still given to @robnasby), so merging #610 would cause some conflicts when you merge this one. If you decide that both #610 and this PR are worth taking, I'd suggest **NOT** merging #610 as-is, since this one replaces it. That way you'll have fewer merge conflicts to deal with.

moh-hassan:
It's a best practice that every PR is one feature at a time per branch. This simplify maintenance and review, even if there is a minor confliction.
#594 and #610 have no breaking change, but #607 have a major breaking change and still need more discussion as I pointed in note 2,3 and 4.
Kindly, read my [comments](https://github.com/commandlineparser/commandline/issues/601#issuecomment-635043612) and comment for note 2,3 and 4.
I take your consideration of #610 and delay approving it.

rmunn:
I've responded to notes 2, 3 and 4 in the other thread. I'll summarize my comments here:

Note 2, about AutoHelp=false: I believe this PR honors AutoHelp=false. If it doesn't, that's a bug and I'll fix it.

Note 3, about `--` being allowed to be used as a value to another option: that should be the default, without any need for a ParserSettings change, because that's how getopt behaves. The fact that CLP currently doesn't allow this is precisely the bug I reported in #601 and that I am fixing here.

Note 4, about allowing an "option" (by which I think you mean a string value starting with `-`, whether or not it's currently a valid option specified by the user) to be used as a value to another option: again, that should be the default, without any need for a ParserSettings change, because that's how getopt behaves. The fact that CLP doesn't allow this is a bug, which I reported in #601 and which I'm fixing here.

BTW, about this being a breaking change: I actually was expecting this to be a breaking change, since I had a big rewrite of the tokenizer. But I was amazed to discover that after I rewrote the tokenizer, all the unit tests still passed! So this isn't a breaking change, just a bugfix. It turns out that after this fix, the behavior of CommandLineParser is *just as it was before*, except that a couple of bugs (`--` not being allowed as a value, for example) are now fixed. And I would be *really* surprised to find that anyone was actually relying on the buggy behavior, because (as you're probably tired of hearing me say by now), the old behavior was not how getopt works, and anyone using this library is probably expecting getopt-like behavior.

So although I won't *object* if you decide that this PR needs a new major version number once it goes in, I do think that a major version number bump is not necessary here, because it's a bugfix that brings the behavior *more* in line with what it's intended to be. It's also not at all a breaking change from the point of view of the API, because the user-facing API has no changes except for a new OptionAttribute property (FlagCounter) being available.

Having said that, I *do* think that more documentation is needed, and once I know you're planning to accept this PR I'll start working on editing the wiki to explain precisely how getopt, and now CLP, handles option values. Because I do think that the documentation is lacking in how it explains getopt's corner cases which CLP mimics, and it might be good to make sure they know what to expect.

moh-hassan:
I'll jump temporary for using #607, for running  sample programs already using v2.8.
I run the program using v2.8 and the Pr607 and compare the result:

Case#1:
I use custom parser with  `EnableDashDash = true;`
I pass plugin options `PluginArgs ` to a third party program:
```cs
  class Options
        {
          // plugin name
            [Option('n')]
            public string Name { get; set; }
          //plugin options 
            [Option('p')]
            public IEnumerable<string> PluginArgs { get; set; }
            
        }
```
commandline: I pass -- so parser can consider all options after -- as a value.

```
$ myprog -n abc -p -- -x -y -z 44"
```
In v2.8
I catch all arguments `-x -y -z 44` correctly in PluginArgs 

in PR607, i get an error
```
DemoClp 1.0.0
Copyright (C) 2020 DemoClp

ERROR(S):
  Option 'x' is unknown.
  Option 'y' is unknown.
  Option 'z' is unknown.

  -n

  -p

  --help       Display this help screen.

  --version    Display version information.
```
I want to migrate the program to PR607 with zero change in the source code to avoid breaking change (if possible).
So, what is the right way to migrate myapp running in v2.8 to work properly in PR607?

rmunn:
That was always an incorrect way to use `--`, and I for one would be filing a bug report if I discovered that as a user I was expected to specify the PluginArgs that way, but I can see your point that some people might be relying on that behavior, so I guess it should be a breaking change after all.

The right way to use a PluginArgs option like that would be to do one of two things:

1. Specify a separator like `' '` (space) and pass the PluginArgs as `-p "-x -y -z 44"` (which was not possible before #607, but is the right way to do it), or

1. Turn on AllowMultiInstance, which then lets the user pass the PluginArgs one at a time such as `-p -x -p -y -p -z -p 44`; again, this would not have been possible before #607 (#594 adds AllowMultiInstance, but #607 allows values that start with `-` to be normal option values without having to misuse `--` for that purpose).

Both do require a small modification to the user's code, so I guess we do have to consider this a breaking change even though it's really a bugfix.

rmunn:
Replaced by #684, which adds a GetoptMode option to the parser settings so that people have to specifically opt-in to this behavior. Closing this PR in favor of #684.

Issue #596, opened 2020-03-12
Parser still recognizes --help option even when i set the parser setting "AutoHelp" to false


elgonzo:
(CommandlineParser version 2.7.82)

When using a parser with the `AutoHelp` setting set to `false` i would expect the parser to not recognize the `--help` option unless i have it explicitly declared as part of my options class. If i have no such declaration, i would expect the parser to treat `--help` as an unknown option/error. Unfortunately, this is not the case. The parser still recognizes this option, even when `AutoHelp` is set to false :(

The [documentation comment](https://github.com/commandlineparser/commandline/blob/017a8a21c79b223fbcd541187c7dd546c18ebdda/src/CommandLine/ParserSettings.cs#L140-L142) for `AutoHelp` (also shown by Intellisense) seems to support my expectations:

> Gets or sets a value indicating whether implicit option or verb 'help' should be supported.

But that does not seem to be what that setting actually does...

Thus my question: What is the actual purpose and supposed effect of setting `AutoHelp` for the parser to `false`? And how would i make `--help` not an option?

For example, if i use the unknown option `--foo`, the `HelpText.AutoBuild` method is able to generate a text containing:
```
ERROR(S):
  Option 'foo' is unknown.
```

I want to get the same result when calling my program with the (unknown) option `--help`:

```
ERROR(S):
  Option 'help' is unknown.
```
Unfortunately, the parser or `HelpText.AutoBuild` method seem to refuse to treat `--help` as an unknown option (both parser settings `AutoHelp` and `IgnoreUnknownArguments` are set to `false`)
:-(

rmunn:
It looks like I'm indirectly responsible for causing this behavior to happen, actually. In https://github.com/gsscoder/commandline/issues/225, I asked for "Option 'help' is unknown" to NOT show up when running a verb-based command like "my-cmd verb arg --help". The solution that was implemented at the time was a band-aid that [filters out UnknownOptionError](https://github.com/commandlineparser/commandline/blob/182e72fa363485f10b2cdbf22b8c1ca23d03eae5/src/CommandLine/ErrorExtensions.cs#L19) if (and only if) the unknown option in question is the "--help" option.

[...]

PR #608, opened 2020-03-27
Fix #600 - better handling of --help and --version options

### LINKED ISSUE #600: Design discussion: how to handle --help option
The design principle behind CommandLineParser is to [mimic the behavior of the standard GNU `getopt` library](https://github.com/commandlineparser/commandline/issues/498#issuecomment-521757260), but the handling of the standard `--help` option is failing to mimic `getopt` correctly.

## How it currently works

Currently, the `--help` option is really only recognized if it's the *first* option passed, e.g.:

* `progname --help` causes the parser to return a HelpRequestedError, which if AutoHelp=true (the default) will print the help text.
* `progname help` causes the parser to return a HelpVerbRequestedError, which if AutoHelp=true will print the help text.
* `progname foo --help` is handled differently depending on whether `foo` is a verb:
    * If `foo` is a verb, then the help for that verb is constructed and returned.
    * If `foo` is NOT a verb, then `progname foo --help` causes the parser to return an UnknownOptionError("help"), which if AutoHelp=true will print the help text, and NOT print "ERROR(S): Option 'help' is unknown." This is because [UnknownOptionError("help") is explicitly filtered out](https://github.com/commandlineparser/commandline/blob/182e72fa363485f10b2cdbf22b8c1ca23d03eae5/src/CommandLine/ErrorExtensions.cs#L19) of the "meaningful errors" list, and only "meaningful errors" are displayed to the user.
* `progname foo bar --help`, when `foo` is a verb, actually does the same as `progname foo --help` when `foo` is not a verb: it returns an UnknownOptionError("help") which is then filtered out of the output.

## How it should work

The [GNU Coding Standards docs on how to handle `--help`](https://www.gnu.org/prep/standards/standards.html#g_t_002d_002dhelp) are short:

> The standard `--help` option should output brief documentation for how to invoke the program, on standard output, then exit successfully. Other options and arguments should be ignored once this is seen, and the program should not perform its normal function. 

This doesn't resolve some of the ambiguous situations. But by experimenting with standard GNU utilities and how they handle the `--help` option, some clear patterns emerge:

* `progname --help` should print help output
* `progname --help foo` should ignore the argument "foo" and print help output.
* `progname arg1 arg2 --help` should ignore `arg1` and `arg2` and print help output. Whether `arg1` and `arg2` were valid options or not should be completely ignored, and the help output should not contain any errors even if `arg1` and `arg2` were invalid.
* `progname --input-file --help`, where `--input-file` is an option that expects a string parameter, is a bit of a gotcha. This should **NOT** print help output; rather, `--input-file` should receive the string `--help` as its parameter. (You can confirm this by running multiple GNU commands on Linux, such as `ls --help` vs `ls --quoting-style --help`, and so on). More about this later.
* `progname verb --help` should display the help text for the specific verb in question *if* verbs are being used, otherwise `verb` should be ignored.
* `progname verb arg1 arg2 --help` should also display the help text for the specific verb in question, and not return any errors about arg1 and arg2.

The GNU standards do not specify whether `-h` should be taken as a short form for the long `--help` option. Here the behavior differs between Linux programs: *some* programs allow `-h` as a short form for `--help`, while others use `-h` as a short form for a different command (e.g., `ls -h` is short for `ls --human-readable`). The way CommandLineParser should handle this is to have a AutoHelpShortName flag (a bool), defaulting to **false*. If it is true, then the `-h` option will be taken as a shortname for `--help`. AutoHelpShortName should *not* be a string or a char; it is *not* recommended to change the shortname of the auto-help parameter from `-h` to something else, so if you want to do that you'll need to set AutoHelp=false and implement your own help option.

### About that `--input-file --help` scenario

The naïve approach, of scanning the incoming arguments to look for a `--help` option, will fail in this scenario. What needs to happen instead is that *all* arguments need to be parsed, and *then* all options (but NOT values) need to be scanned for the `--help` option, and/or the `-h` option if AutoHelpShortName is true. If the `--help` option is found, then all other errors are *ignored* (as per [the GNU docs](https://www.gnu.org/prep/standards/standards.html#g_t_002d_002dhelp)) and the help text is printed out. (The only time errors should be printed out is if `--help` is not present).

## What about `--version`?

The [GNU docs for `--version`](https://www.gnu.org/prep/standards/standards.html#g_t_002d_002dversion) are similar to `--help`:

> The standard `--version` option should direct the program to print information about its name, version, origin and legal status, all on standard output, and then exit successfully. Other options and arguments should be ignored once this is seen, and the program should not perform its normal function.

They go on to explain in more detail how to format the version information, but that's not really relevant for this issue. What's important here is that `--version` should be treated just like `--help`: it should handle `--version` appearing in *any* position in the options (except as a value to some other option), and should not return any other errors if `--version` is found. Also, there should be flag named AutoVersionShortName, defaulting to false, which will use the shortname `-V` (capital V) if it is true. (Lowercase `-v` is usually reserved for `--verbose`; the de facto standard in Linux utilities is that if `--version` has a shortname, it's capital `-V`).

## PR forthcoming

I plan to write a PR to handle `--help` and `--version` correctly, as opposed to the slightly-wrong way that CommandLineParser currently handles it. Expect that PR in the next few weeks.

rmunn:
This PR builds on #607 to fix #600. Now that string values are handled by the tokenizer exactly how getopt handles them, we can also handle the special `--`, `--help` and `--version` options exactly how getopt handles them. In particular, we can respond to `--help` no matter where it appears in the argument list (the previous behavior would only respond to `--help` if it was the first parameter, and had some hacks that tried to ensure that it would be responded to elsewhere, but those hacks were incomplete and buggy, creating issues like #596 among others).

Note that because I wrote this PR on top of #607, merging this PR will automatically merge #607 as well. I did it this way because proper tokenization of `--help` (and NOT handling it if it was a string value to some other option) is essential to how this PR is supposed to work. Without #607, many of the tests I wrote for the AutoHelp and AutoVersion options would fail.

### DIFF
--- a/src/CommandLine/Core/ArgumentsExtensions.cs
+++ b/src/CommandLine/Core/ArgumentsExtensions.cs
@@ -10,9 +10,9 @@ namespace CommandLine.Core
     static class ArgumentsExtensions
     {
         public static IEnumerable<Error> Preprocess(
-            this IEnumerable<string> arguments,
+            this IEnumerable<Token> arguments,
             IEnumerable<
-                    Func<IEnumerable<string>, IEnumerable<Error>>
+                    Func<IEnumerable<Token>, IEnumerable<Error>>
                 > preprocessorLookup)
         {
             return preprocessorLookup.TryHead().MapValueOrDefault(
--- a/src/CommandLine/Core/InstanceBuilder.cs
+++ b/src/CommandLine/Core/InstanceBuilder.cs
@@ -32,7 +32,9 @@ namespace CommandLine.Core
                 ignoreValueCase,
                 parsingCulture,
                 autoHelp,
+                false,
                 autoVersion,
+                false,
                 false,
                 nonFatalErrors);
         }
@@ -45,7 +47,9 @@ namespace CommandLine.Core
             bool ignoreValueCase,
             CultureInfo parsingCulture,
             bool autoHelp,
+            bool autoHelpShortName,
             bool autoVersion,
+            bool autoVersionShortName,
             bool allowMultiInstance,
             IEnumerable<ErrorType> nonFatalErrors)
         {
--- a/src/CommandLine/Core/InstanceBuilder.cs
+++ b/src/CommandLine/Core/InstanceBuilder.cs
@@ -61,7 +61,9 @@ namespace CommandLine.Core
 
             var specs = from pt in specProps select pt.Specification;
 
-            var optionSpecs = specs
+            var autoSpecs = AddAutoSpecs(specs, nameComparer, autoHelp, autoHelpShortName, autoVersion, autoVersionShortName);
+
+            var optionSpecs = autoSpecs
                 .ThrowingValidate(SpecificationGuards.Lookup)
                 .OfType<OptionSpecification>()
                 .Memoize();
@@ -76,12 +78,13 @@ namespace CommandLine.Core
                 errs => new NotParsed<T>(makeDefault().GetType().ToTypeInfo(), errs);
 
             var argumentsList = arguments.Memoize();
+
+            var tokenizerResult = tokenizer(argumentsList, optionSpecs);
+
+            var tokens = tokenizerResult.SucceededWith().Memoize();
+
             Func<ParserResult<T>> buildUp = () =>
             {
-                var tokenizerResult = tokenizer(argumentsList, optionSpecs);
-
-                var tokens = tokenizerResult.SucceededWith().Memoize();
-
                 var partitions = TokenPartitioner.Partition(
                     tokens,
                     name => TypeLookup.FindTypeDescriptorAndSibling(name, optionSpecs, nameComparer));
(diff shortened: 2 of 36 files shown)


rmunn:
And because #607 is built on top of #594, merging this PR will merge not only #607 but #594 as well.

rmunn:
Now that I've incorporated #610 into #607, I'll be rebasing this PR on top of the new version of #607. Please do not merge until I've done that.

rmunn:
I've rebased this PR on top of the new version of #607. It's ready to merge now.

rmunn:
Rebasing again since #607 has had some changes, in particular it now has a working implementation (and a unit test) for the FlagCounter attribute added to Options.

rmunn:
Replaced by #686, which separates out the `--help` and `--version` handling into its own single PR instead of having a mixed PR like this one. Closing this PR in favor of #686.

Issue #606, opened 2020-03-26
System.ArgumentOutOfRangeException in quick start example ( kind of )


FelicePollano:
Created a really simple app, with option defined like this:
```
public class Options
    {
        
        [Option('a',"address",Default="http://localhost:9200",HelpText="......")]
        public string Address { get; set; }
        [Option('c',"config",Default="elasticizerfile",HelpText=".......")]
        public string Config{get;set;}
         [Option]
        public IEnumerable<string> Files { get;set; }
    }
```

by using the code:
`Parser.Default.ParseArguments(args).WithParsed<Options>(o => {});`
It raises the ArgumentOutOfRange exception.

FelicePollano:
Sorry I was just wrong in calling, correct way is:
Parser.Default.ParseArguments<Options>(args).WithParsed<Options>(o => {});

PR #609, opened 2020-03-27
fix issue #606 (#606)

FelicePollano:
I think checking for types count > 0 is an error since they are optional parameters...

 `public ParserResult<object> ParseArguments(IEnumerable<string> args, params Type[] types)`


FelicePollano:
Please ignore, was a wrong fix

PR #616, opened 2020-04-14
Add generalized implementation of MapResult which supports infinite p…

sin0emp:
As `MapResult` was not supporting more than 16 verbs in its generic templates, I patched this which supports infinite number of verbs. I think it can be a more generalized implementation.

### DIFF
--- a/src/CommandLine/ParserResultExtensions.cs
+++ b/src/CommandLine/ParserResultExtensions.cs
@@ -2,6 +2,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Linq;
 
 namespace CommandLine
 {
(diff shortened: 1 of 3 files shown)


moh-hassan:
Thanks for work.
Can you add #621 as a unit test?

sin0emp:
@moh-hassan it's done now

moh-hassan:
Can you show an example how to use this method in case of async/await?
It's nice if you add it as a test case.

moh-hassan:
I did a test for async/await and it's fine.
We can do it in wiki, and no need to modify PR.

moh-hassan (approved):

sin0emp:
fixed a wrong mapping value in tests

PR #625, opened 2020-05-11
Add multi-instance option support globally and per option

geckosoft:
This PR continues on the work done in PR #594 and makes the following changes:

- Renames the parser setting to AllowMultiInstanceByDefault
- Add a per-option AllowMultiInstance option that can be set in the constructor (by default, each option looks at the AllowMultiInstanceByDefault parser setting)
- Added 2 additional tests (1 for the AllowMultiInstance option)
- Previously enabling AllowMultiInstance(ByDefault) would disable any 'EnforceSingle' check, now it will perform all checks again but at the final step we look at the AllowMultiInstanceByDefault parser setting and the AllowMultiInstance option

Any feedback is welcome.

### DIFF
--- a/src/CommandLine/Core/InstanceBuilder.cs
+++ b/src/CommandLine/Core/InstanceBuilder.cs
@@ -24,6 +24,31 @@ namespace CommandLine.Core
             bool autoVersion,
             IEnumerable<ErrorType> nonFatalErrors)
         {
+            return Build(
+                factory,
+                tokenizer,
+                arguments,
+                nameComparer,
+                ignoreValueCase,
+                parsingCulture,
+                autoHelp,
+                autoVersion,
+                false,
+                nonFatalErrors);
+        }
+
+        public static ParserResult<T> Build<T>(
+            Maybe<Func<T>> factory,
+            Func<IEnumerable<string>, IEnumerable<OptionSpecification>, Result<IEnumerable<Token>, Error>> tokenizer,
+            IEnumerable<string> arguments,
+            StringComparer nameComparer,
+            bool ignoreValueCase,
+            CultureInfo parsingCulture,
+            bool autoHelp,
+            bool autoVersion,
+            bool allowMultiInstance,
+            IEnumerable<ErrorType> nonFatalErrors)
+        {
             var typeInfo = factory.MapValueOrDefault(f => f().GetType(), typeof(T));
 
             var specProps = typeInfo.GetSpecifications(pi => SpecificationProperty.Create(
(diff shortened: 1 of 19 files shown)


rmunn:
While I'm not the maintainer, nullable boolean arguments don't fit with the code style. If you really have to have a three-valued bool (true, false, or not specified) for allowMultiInstance in OptionSpecification, I'd suggest using a `Maybe<bool>` instead and the appropriate Maybe-handling functions from Csharpx, as found in many other parts of the code. But my personal preference would be to just say "not specified = false" and use a regular boolean.

mylemans:
Doing the not specified = false would make it impossible to enable multi instance by default and opt-out per option though.

Also, if I read Maybe correctly, if I'd use Maybe.Return(false) it would actually return an instance of Nothing<bool>

rmunn:
Ah, right. Then a `Maybe<bool>` is the right approach here. And no, `Maybe.Return(false)` should not return a `Nothing` instance, at least if it's implemented properly. I'll check.

rmunn:
Okay, looks like there's a bug in Maybe. The `Return` function is *supposed* to always return a `Just` instance instead of `Nothing`, but the way it was implemented, you're right that it would turn a `false` into a `Nothing`, incorrectly. To work around that, use `Maybe.Just` instead of `Maybe.Return` when you know the value *should* be `Just(false)`.

mylemans:
I'll see if I can make the requested changes. I'm not the biggest fan of using Maybe for this, but I'll see if I can comply.

mylemans:
@rmunn Hmm, Maybe isn't public so I cannot use this type for the public AllowMultiInstance read-only option on OptionAttribute, should I make that property internal?

edit: Maybe I should retain the nullable bool there and change everywhere else to use this 'Maybe'? Like in FromAttribute and OptionSpecification

rmunn:
If it's a property on OptionAttribute, then it has to be public, since that's what library users will be accessing when they write `Option("foo", AllowMultiInstance=false)`. BTW, I believe it should be a read-write property so that library users can set it on the options, like with the Min and Max properties in BaseAttribute. And yes, perhaps a nullable bool on the allowMultiInstance *member* of OptionAttribute would fly. For the *property*, though, I suggest that it should be a simple bool, whose getter should calculate the effective value (based on whether it was set or not, and based on the default setting if it wasn't set) and whose setter should set the member.

mylemans:
AllowMultiInstance cannot be a nullable bool then though (not allowed in attributes).

Making it only set the nullable bool internally on set could work (be it false or true) but only for writing.

For calculating the effective value, you will need the current parser' context to get that setting, not sure how you would feed that info to the option?

PR #634, opened 2020-05-22
Expose value and error on parser result

### LINKED ISSUE #165: Consider alternative syntax
<a href="https://github.com/Korporal"></img></a> **Issue by [Korporal](https://github.com/Korporal)**
_Monday Sep 04, 2017 at 15:28 GMT_
_Originally opened as https://github.com/gsscoder/commandline/issues/476_

----

Hi,

Consider this generic approach, I think it's better for developers, simpler pattern:

```
static void Main(string[] args) {
  
    var options = CommandLine.Parser.Default.ParseArguments<Options>(args);

    if (options.Available)
       // Values are available here
       if (options.Verbose) Console.WriteLine("Filename: {0}", options.InputFile);
 
}
```

just a thought...

### LINKED ISSUE #543: Expose parsed options object in ParserResult
I tried using this library in my project today. I'm able to parse the command line from the examples just fine, but found no way to access the parsed options object directly from `ParserResult`.

By my coding style, I prefer not use callbacks here (i.e. `WithParsed` and `WithNotParsed`). I'm already able to use the `Tag` property to check if parsing succeeded. Am I just missing something here or can a property exposing the parsed object be added?

johnjaylward:
Fixes #543
Fixes #165

@moh-hassan no existing tests break when I run them through visual studio. I have not added any new tests. Do you want to see any specific examples be tested?

### DIFF
--- a/src/CommandLine/ParserResult.cs
+++ b/src/CommandLine/ParserResult.cs
@@ -67,17 +67,17 @@ namespace CommandLine
         internal ParserResult(IEnumerable<Error> errors, TypeInfo typeInfo)
         {
             this.tag = ParserResultType.NotParsed;
-            this.typeInfo = typeInfo;
-            Errors = errors;
+            this.typeInfo = typeInfo ?? TypeInfo.Create(typeof(T));
+            Errors = errors ?? new Error[0];
             Value = default;
         }
 
         internal ParserResult(T value, TypeInfo typeInfo)
         {
+            Value = value ?? throw new ArgumentNullException(nameof(value));
             this.tag = ParserResultType.Parsed;
-            this.typeInfo = typeInfo;
+            this.typeInfo = typeInfo ?? TypeInfo.Create(value.GetType());
             Errors = new Error[0];
-            Value = value;
         }
 
         /// <summary>
--- a/src/CommandLine/ParserResult.cs
+++ b/src/CommandLine/ParserResult.cs
@@ -9,7 +9,7 @@ namespace CommandLine
     public sealed class TypeInfo
     {
         private readonly Type current;
-        private readonly IEnumerable<Type> choices; 
+        private readonly IEnumerable<Type> choices;
 
         private TypeInfo(Type current, IEnumerable<Type> choices)
         {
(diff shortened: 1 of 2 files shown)


src/CommandLine/ParserResult.cs
```diff
@@ -159,21 +168,12 @@ public bool Equals(Parsed<T> other)
     /// <typeparam name="T">The type with attributes that define the syntax of parsing rules.</typeparam>
     public sealed class NotParsed<T> : ParserResult<T>, IEquatable<NotParsed<T>>
     {
-        private readonly IEnumerable<Error> errors;
 
-        internal NotParsed(TypeInfo typeInfo, IEnumerable<Error> errors)
-            : base(ParserResultType.NotParsed, typeInfo)
+        internal NotParsed(IEnumerable<Error> errors, TypeInfo typeInfo)
```
moh-hassan: Is there a reason to switch the parameters of ctor? It's better to keep internal API the same.
It also keep tests without change.

johnjaylward: I was trying to be have the API be consistent across the Hierarchy.
`ParseResult(value/error, type)` matches `Parsed(value, type)` and `NotParsed(error, type)`

I thought it made it easier to see the relation between the constructors better, but I can revert it if you prefer having the parameter orders not match.

moh-hassan:
Good solution.
Please, can you add a test case for simple option and simple two verbs to test the new feature.
These tests can be used in wiki.

src/CommandLine/ParserResult.cs
moh-hassan: I see, but keeping test without change is very necessary. It proves no break change.

johnjaylward: I'm not sure I understand your comment. Are you saying that there may be other places using the "internal" constructor outside of this library? That shouldn't be possible unless someone is doing very weird reflection:

https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/internal

the change should only affect the internals of this library, which includes the test cases.

moh-hassan: Internals are used by other PRs and test suit.
There are some PRs are still open and sure that will effect their work when merged and cause confliction when merging.

shravan2x:
@johnjaylward Thank you so much for making this PR! The old flow felt very awkward and this change makes my code cleaner :)

src/CommandLine/ParserResult.cs
johnjaylward: I will revert those changes then.

moh-hassan: Thanks for the good work.

johnjaylward:
I added the test cases. Let me know if they match your expectations or need more improvements.

moh-hassan:
Yes, the tests are enough and no breaking change :+1: 
Congratulation.

moh-hassan (approved):

johnjaylward:
fixes #543

PR #642, opened 2020-06-04
Bugfix: Binding Localized Resource to Argument Help Text

nmchapin:
This patch fixes the following issue:

* When specifying a name of a resource to use for the help text of an argument, the utility method responsible for retrieving the resource fails. One of the binding flags in this method (BindingFlag.GetProperty) is invalid. It should only be used in conjunction with the Type.InvokeMember method.

* The exception message that is thrown when a resource cannot be found will now contain the name of the resource in question. The exception message string was obviously supposed to be interpolated but the dollar sign marking it as such was missing.


moh-hassan:
Can you show by example current and suggested solution.

PR #636, opened 2020-05-26
Verb aliases

johnjaylward:
Fixes #6 
Fixes #517

PR #650, opened 2020-06-18
fix spacing on help messages when verbs have aliases

### LINKED ISSUE #6: Verb aliases
<a href="https://github.com/kirillkovalenko"></img></a> **Issue by [kirillkovalenko](https://github.com/kirillkovalenko)**
_Tuesday Aug 20, 2013 at 17:19 GMT_
_Originally opened as https://github.com/gsscoder/commandline/issues/90_

----

I suggest implementing verb aliases. For example: hg rename, hg move, hg mv is the same command.

### LINKED ISSUE #517: Improvement: short form for verbs
We need to have some way to specify short version of the verb string, like

[Verb("inventory", alias="inv" HelpText="Show the Inventory")]
class ....

johnjaylward:
Fixes #6 
Fixes #517 
Fixes #636

Uses suggestion from @twaalewijn in https://github.com/commandlineparser/commandline/pull/636#issuecomment-645501830 to fix layout of help messages. Uses existing extension method for the concat instead of the exact one provided in the example.

### DIFF
--- a/src/CommandLine/Text/HelpText.cs
+++ b/src/CommandLine/Text/HelpText.cs
@@ -851,8 +851,8 @@ namespace CommandLine.Text
             var optionSpecs = from verbTuple in Verb.SelectFromTypes(types)
                               select
                                   OptionSpecification.NewSwitch(
-                                      verbTuple.Item1.Name,
-                                      verbTuple.Item1.Aliases.ToDelimitedString(", "),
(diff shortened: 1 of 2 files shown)


twaalewijn:
Can confirm that the fix in this PR also causes the tests that I added in my branch to succeed again. 👍

PR #658, opened 2020-06-30
Add the simple HelpText.AutoBuild() function

loxsmoke:
This is the minimal AutoText() function that can generate help text only from the command line parameters class. 

Example:
```
HelpText helpText = HelpText.AutoBuild<MyCommandLine>();
Console.WriteLine(helpText.ToString());
```

I find it very useful for simple utilities that do not use verbs and have to show help text in case of invalid input data that cannot be detected by parsing.


bannikovilea:
Hi!) I would be glad to get this method for my small project. Can you merge it ?

PR #669, opened 2020-07-24
Allow single dash as a value

### LINKED ISSUE #574: What is the correct way to parse an argument with value -
I have a program that I would like to execute like so:

`./myprogram --someArg 2 --thisFails - --anotherArg 3`

Where the value of `thisFails` should be set to `-`

I've also tried: `"-"` to no avail

The error I'm getting is:

```
ERROR(S):
  Option 'thisFails' has no value.
```

moh-hassan:
Fix #574
Avoid using triple quote in a single dash """-""" to be parsed as a value.

### DIFF
--- a/src/CommandLine/Core/Tokenizer.cs
+++ b/src/CommandLine/Core/Tokenizer.cs
@@ -135,6 +135,12 @@ namespace CommandLine.Core
             string value,
             Func<string, NameLookupResult> nameLookup)
         {
+            //Allow single dash as a value
+            if (value.Length == 1 && value[0] == '-')
+            {
+                yield return Token.Value(value);
+                yield break;
+            }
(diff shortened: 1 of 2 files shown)


moh-hassan:
@gsscoder 
@ericnewton76 
Please, update the key of nuget for one year to enable publishing nuget package v2.9.1.
Two build failed, [2.9.0-preview2](https://ci.appveyor.com/project/commandlineparser/commandline/builds/34725712/job/06ddhkp457929o26) and [2.9.0](https://ci.appveyor.com/project/commandlineparser/commandline/builds/34513324/job/4wven0308etpvtav), with error message:
>Error publishing package. NuGet server returned 403: The specified API key is invalid, has expired, or does not have permission to access the specified package

Issue #396, opened 2019-01-22
parsing fails when value matches name of list parameter


vemoo:
```csharp
class Options
{
    [Option("deploy", Separator = ',', = "Projects to deploy")]
    public IList<string> Deploys { get; set; }

    [Option("profile", Required = true, HelpText = "Profile to use when restoring and publishing")]
    public string Profile { get; set; }

    [Option("configure-profile", Required = true, HelpText = "Profile to use for Configure")]
    public string ConfigureProfile { get; set; }
}
```
Parsing fails if passed `--configure-profile deploy --profile local` with `Required option 'profile' is missing.`
I think it's intepreting the `deploy` value as an option.

moh-hassan:
@rmunn 
This is a known issue in the library, see this [test case](https://dotnetfiddle.net/eTyrjR).
Please, can you help in fixing this issue?

rmunn:
@moh-hassan Interesting. I'll take a look.

rmunn:
Initial findings: the tokenizer returned a Name token for `--configure-profile`, followed by three Value tokens for `deploy`, `--profile` (which is wrong, should have been a Name token) and `local`.

But I just stepped through the tokenizer and watched it happen, and the tokenizer initially returned a Name token for `--profile`. But then inside the Tokenizer.ExplodeOptionList method, that Name token was turned into a Value token. So the ExplodeOptionList method is the culprit. I'll continue tracking it down and report more.

BTW, my GetoptTokenizer implementation from PR #684 also contains this bug, because I re-used that ExplodeOptionList method without any changes. So when I find and fix this bug, I'll also submit an update to PR #684. Please don't merge #684 until then, or you'll be merging known-buggy code that I'm working on a fix for.

rmunn:
Related, though not the same bug: #687 (which I'll also submit a fix for).

rmunn:
PR #684 now has a fix for this bug (in commit 47618e0ddf43a380baaa54484b188118b30d8a6c).

PR #672, opened 2020-07-28
Fix #420 - Check for value token before expanding enumerable types with separator char

### LINKED ISSUE #420: Parsing issues when option has separator attribute and when empty value provided... fails
```
 [Option('b', "BookIds", Required = false, HelpText = "List of Books", Separator = ',')]
        public IEnumerable<string> Bookids { get; set; }

[Option('v', "Verbose", Required = false, Hidden = true, HelpText = "Generate Book details", Default = "Yes")]
        public string All { get; set; }
```
ProcessBooks -b -v NO

------------
If option has separator attribute and when empty value provided... fails; it considers other argument as its values..

In this case; -b has value -v and NO;

wo-roberts:


rmunn:
I don't think this is quite the right solution. I think in addition to checking if the token to be exploded is a Value, we *also* need to check if the previous token is a Name. I'll try to create a unit test that proves that.

BTW, I'm currently working on a fix for https://github.com/commandlineparser/commandline/issues/396 that looks like it's going to end up fixing #420 at the same time, so once that PR is submitted I'll link to it from this one as well.

rmunn:
Here's a unit test that fails with this PR, but works if you also check that the previous token is a Name:

```csharp
[Fact]
public static void Values_with_same_name_as_sequence_option_do_not_cause_later_values_to_split_on_separators()
{
    var args = new[] { "c", "x,y" };
    var tokensExpected = new[] { Token.Value("c"), Token.Value("x,y") };
    var typeInfo = typeof(OptionsWithSeparator);

    var specProps = typeInfo.GetSpecifications(pi => SpecificationProperty.Create(
            Specification.FromProperty(pi), pi, Maybe.Nothing<object>()))
        .Select(sp => sp.Specification)
        .OfType<OptionSpecification>();

    var tokenizerResult = Tokenizer.ConfigureTokenizer(StringComparer.InvariantCulture, false, false)(args, specProps);
    var tokens = tokenizerResult.SucceededWith();
    tokens.Should().BeEquivalentTo(tokensExpected);
}

internal class OptionsWithSeparator
{
        [Option('f', "flag", HelpText = "Flag")]
        public bool Flag { get; set; }

        [Option('c', "categories", Required = false, Separator = ',', HelpText = "Categories")]
        public IEnumerable<string> Categories { get; set; }

        [Option('j', "jobId", Required = true, HelpText = "Texts.ExplainJob")]
        public int JobId { get; set; }
}
```

rmunn:
https://github.com/commandlineparser/commandline/pull/684/commits/47618e0ddf43a380baaa54484b188118b30d8a6c (part of #684) is a better fix, which fixes the unit test in https://github.com/commandlineparser/commandline/pull/672#issuecomment-677451139 that fails with this PR.

PR #683, opened 2020-08-19
More tests for double-dash scenarios

rmunn:
More unit tests to verify various combinations of a `--` arg in the middle of a sequence. These are a few tests that I meant to include in #678 but accidentally left out.


ericnewton76 (commented):
~~Sorry but none of these tests make sense to me~~

Ok I re-acquainted myself with `--` and EnableDashDash

PR #684, opened 2020-08-19
Add GetoptMode parser setting and implementation

### LINKED ISSUE #91: MissingRequiredOptionError when specifying required Value after enumerable Options
<a href="https://github.com/SuperFXMaster"></img></a> **Issue by [SuperFXMaster](https://github.com/SuperFXMaster)**
_Tuesday Oct 11, 2016 at 19:24 GMT_
_Originally opened as https://github.com/gsscoder/commandline/issues/361_

----

The following options class:

``` C#
public class CommandLineOptions
{
    [Value(0, Required = true)]
    public string InputFileName { get; set; }

    [Option('o', "output")]
    public string OutputFileName { get; set; }

    [Option('i', "include", Separator = ',')]
    public IEnumerable<string> Included { get; set; }

    [Option('e', "exclude", Separator = ',')]
    public IEnumerable<string> Excluded { get; set; }
}
```

When provided to the default parser, with the commandline `app.exe --exclude=a,b InputFile.txt` results in a `MissingRequiredOptionError: A required value not bound to option name is missing`. I expected the library to accept `InputFileName = InputFile.txt` and `--exclude=a,b` as `Excluded = { "a", "b" }`. Is there something else required to make this work as I expect, or is it a bug?

### LINKED ISSUE #396: parsing fails when value matches name of list parameter

### LINKED ISSUE #420: Parsing issues when option has separator attribute and when empty value provided... fails

### LINKED ISSUE #454: Wrong management of sequences and values
Consider following definition of options:
```
    class Options
    {
        [Option('c', "channels", Required = true, Separator = ':', HelpText = "Channel names")]
        public IEnumerable<string> Channels { get; set; }

        [Value(0,
               Required = true,
               MetaName = "file_path",
               HelpText = "Path of archive to be processed")]
        public string ArchivePath { get; set; }
    }
```
When invoking `app.exe -c chanA:chanB file.hdf5` the expected result would be 
```
Channels = {"chanA", "chanB"} 
ArchivePath = "file.hdf5"
```
Instead, I get a `CommandLine.MissingRequiredOptionError`

If I remove the `Required` attribute from the value definition, parsing succeeds but I get:
`Channels = {"chanA", "chanB", "file.hdf5"} `
And the expected `ArchivePath `value ends up in the `Channels `definitions (why? there is no ":" separator)

Am I doing something wrong?

### LINKED ISSUE #510: Using IEnumerable for Option is always 'Required'
[Option('a', "aa", Required = false, Separator = ',')]
        public IEnumerable<string> A { get; set; }
[Option('b', "bb", Required = false)]
        public string B { get; set; }
[Value(0, Required = true)]
        public string C { get; set; }

For some reasons, if I only specify Option A and C, I will get the following error:

ERROR(S):
  A required value not bound to option name is missing.
  Error setting value to option 'a, aa': Check if Option or Value attribute values are set properly for the given type.

### LINKED ISSUE #601: Corner cases where getopt behavior is not mimiced: -- or --help as string values

### LINKED ISSUE #617: Enumerable of Enum doesn't work if followed by Values
I have such options:
```csharp
[Option("fm",  Separator=',', Default = new[] { FMode.S })]  
public IEnumerable<FMode> Mode { get; set; }
	 
[Option('q')]
public bool q { get;set; }
	 
[Value(0)]
public IList<string> Files { get; set; }
```

If I run my application with `--fm D,C a.txt` arguments, it says that option 'fm' is defined with a bad format. If I put any option between 'fm' and the value list, for example, `--fm D,C -q a.txt`, it is parsed successfully. You can check it in a fiddle: https://dotnetfiddle.net/yt5L9K

### LINKED ISSUE #619: Order of options where there are list options and values
Hi.

The attached file contains the configuration to parse a number of options that consist of list of elements and finally a list of filenames (after the --).
The problem appear when the last option prior to the -- is a list based option. In that case the parser responds with: 'A required value not bound to option name is missing'

The following command line work with the provided C# configuration:
--modules ../utilities/x64/Debug,../auxtool/x64/Debug --outdir ./x64/Debug -- m_xfunit.f03 m_xfunit_assertion.f03

The following command line does not (with the error mentioned above)
--outdir ./x64/Debug --modules ../utilities/x64/Debug,../auxtool/x64/Debug -- m_xfunit.f03 m_xfunit_assertion.f03

Am I overlooking something or is this an issue or limitation in the software?

Regards,
FranHi.

The attached file contains the configuration to parse a number of options that consist of list of elements and finally a list of filenames (after the --).
The problem appear when the last option prior to the -- is a list based option. In that case the parser responds with: 'A required value not bound to option name is missing'

The following command line work with the provided C# configuration:
--modules ../utilities/x64/Debug,../auxtool/x64/Debug --outdir ./x64/Debug -- m_xfunit.f03 m_xfunit_assertion.f03

The following command line does not (with the error mentioned above)
--outdir ./x64/Debug --modules ../utilities/x64/Debug,../auxtool/x64/Debug -- m_xfunit.f03 m_xfunit_assertion.f03

Am I overlooking something or is this an issue or limitation in the software?

Regards,
Fran
[CommandLineOptions.cs.txt](https://github.com/commandlineparser/commandline/files/4508735/CommandLineOptions.cs.txt)

### LINKED ISSUE #631: Is it possible to use repeated non-option-argument options?
ssh, tar, and other common tools allow for multiple non-option-argument options and act on their total count. (Implementing this with getopt is pretty trivial).

e.g. `tar -cvf` and `tar -cvvf` produce different output.

Is there a way to do this with CommandLine? I did not see anything promising in the documentation or examples/source code I perused, but I may have missed something.

(This might just be related to #594, but the issue there seems to be with repeated options with _required_ option-arguments)

### LINKED ISSUE #687: Sequences with separators, if a separator is provided on the command line, should STOP at the separator
Consider this Options class:

```csharp
public class OptionsWithSeparator
{
        [Option('c', "categories", Separator = ',', HelpText = "Categories")]
        public IEnumerable<string> Categories { get; set; }

        [Value(0, HelpText = "File name")]
        public string Filename { get; set; }
}
```

What would you expect to get from the commandline `myprog -c one,two filename.txt`?

If you expect that to produce `Categories = ["one", "two"], Filename = "filename.txt"`, I agree. But that's not what happens. What it currently produces is `Categories = ["one", "two", "filename.txt"], Filename = null`. This is caused by the fact that IEnumerable options try to grab every subsequent non-option argument that follows them, and only stops once another option has been encountered (or, if EnableDashDash is true and you're using the upcoming 2.9 release, once a `--` argument has been encountered).

I believe that this is unintuitive behavior, and that what *should* happen with IEnumerable options is the following:

* If there is no separator given, grab all available non-option args
* If a separator is given, but the next non-option arg doesn't contain the separator, continue to grab all available non-option args
* If a separator is given, and the next non-option arg contains the separator, STOP and don't grab any further values (**this would be new behavior**)

Currently the code follows the first two of those rules, but not the third.

Here are bug reports that are ultimately caused by this overly-greedy grabbing of IEnumerables, which would have been solved by the behavior I listed above:

#91
#454
#510
#617
#619 (which can also be fixed by setting EnableDashDash = true, but it would *also* be fixed by the behavior I'm proposing here)

Clearly, users expect a separator, if provided, to confine the IEnumerable's "grabbing" to just the parameter containing the separator. I'll submit a PR soon to implement this behavior.

rmunn:
Turning on Getopt mode automatically turns on the EnableDashDash and AllowMultiInstance settings as well, but they can be disabled by explicitly setting them to false in the parser settings. This is essentially #607 rebased onto the develop branch, but with an option (defaulting to false) to turn it on. If the user doesn't set GetoptMode to true, then no behavior changes.

Note that two unit tests will fail until https://github.com/commandlineparser/commandline/pull/682 is merged, because they test for the same bug that #682 fixes.

This PR includes the FlagCounter feature from #607 as well; I've made the FlagCounter feature and the GetoptMode feature two separate commits so that if you want to use one but not the other, it's easy to cherry-pick the one you want to use. They don't depend on each other; in particular, FlagCounter works even without GetoptMode being true. But I've included both of them in this PR because GetoptMode without FlagCounter feels incomplete: people running in Getopt mode are going to expect to be able to do `-vvv` and get VerboseLevel=3.

My next PR will be a rebasing of #608 onto develop. Once that's done, if I'm sure that this PR is going to be merged, I'll start working on writing documentation of GetoptMode for the CLP wiki. In the meantime, the unit tests in GetoptParserTests.cs are pretty self-documenting.

Fixes #601.
Fixes #631.

I've also fixed #687 in this PR (in commit https://github.com/commandlineparser/commandline/pull/684/commits/47618e0ddf43a380baaa54484b188118b30d8a6c); I chose to incorporate that fix into this PR because GetoptTokenizer, introduced here, contains the same buggy ExplodeOptionList method as the original Tokenizer, and both of them need to be fixed to properly fix #687. In addition to #687, this PR now:

Fixes #619
Fixes #617
Fixes #510
Fixes #454
Fixes #420
Fixes #396
Fixes #91

### DIFF
--- a/src/CommandLine/Core/GetoptTokenizer.cs
+++ b/src/CommandLine/Core/GetoptTokenizer.cs
@@ -94,19 +94,28 @@ namespace CommandLine.Core
         {
             var tokens = tokenizerResult.SucceededWith().Memoize();
 
-            var replaces = tokens.Select((t, i) =>
-                optionSequenceWithSeparatorLookup(t.Text)
-                    .MapValueOrDefault(sep => Tuple.Create(i + 1, sep),
-                        Tuple.Create(-1, '\0'))).SkipWhile(x => x.Item1 < 0).Memoize();
-
-            var exploded = tokens.Select((t, i) =>
-                        replaces.FirstOrDefault(x => x.Item1 == i).ToMaybe()
-                            .MapValueOrDefault(r => t.Text.Split(r.Item2).Select(Token.Value),
-                                Enumerable.Empty<Token>().Concat(new[] { t })));
-
-            var flattened = exploded.SelectMany(x => x);
-
-            return Result.Succeed(flattened, tokenizerResult.SuccessMessages());
+            var exploded = new List<Token>(tokens is ICollection<Token> coll ? coll.Count : tokens.Count());
+            var nothing = Maybe.Nothing<char>();  // Re-use same Nothing instance for efficiency
+            var separator = nothing;
+            foreach (var token in tokens) {
+                if (token.IsName()) {
+                    separator = optionSequenceWithSeparatorLookup(token.Text);
+                    exploded.Add(token);
+                } else {
+                    // Forced values are never considered option values, so they should not be split
+                    if (separator.MatchJust(out char sep) && sep != '\0' && !token.IsValueForced()) {
+                        if (token.Text.Contains(sep)) {
+                            exploded.AddRange(token.Text.Split(sep).Select(Token.ValueFromSeparator));
+                        } else {
+                            exploded.Add(token);
+                        }
+                    } else {
+                        exploded.Add(token);
+                    }
+                    separator = nothing;  // Only first value after a separator can possibly be split
+                }
+            }
+            return Result.Succeed(exploded as IEnumerable<Token>, tokenizerResult.SuccessMessages());
         }
 
         public static Func<
--- a/src/CommandLine/Core/GetoptTokenizer.cs
+++ b/src/CommandLine/Core/GetoptTokenizer.cs
@@ -1,0 +1,219 @@
+﻿// Copyright 2005-2015 Giacomo Stelluti Scala & Contributors. All rights reserved. See License.md in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using CommandLine.Infrastructure;
+using CSharpx;
+using RailwaySharp.ErrorHandling;
+using System.Text.RegularExpressions;
+
+namespace CommandLine.Core
+{
+    static class GetoptTokenizer
+    {
+        public static Result<IEnumerable<Token>, Error> Tokenize(
+            IEnumerable<string> arguments,
+            Func<string, NameLookupResult> nameLookup)
+        {
+            return GetoptTokenizer.Tokenize(arguments, nameLookup, ignoreUnknownArguments:false, allowDashDash:true, posixlyCorrect:false);
+        }
+
+        public static Result<IEnumerable<Token>, Error> Tokenize(
+            IEnumerable<string> arguments,
+            Func<string, NameLookupResult> nameLookup,
+            bool ignoreUnknownArguments,
+            bool allowDashDash,
+            bool posixlyCorrect)
+        {
+            var errors = new List<Error>();
+            Action<string> onBadFormatToken = arg => errors.Add(new BadFormatTokenError(arg));
+            Action<string> unknownOptionError = name => errors.Add(new UnknownOptionError(name));
+            Action<string> doNothing = name => {};
+            Action<string> onUnknownOption = ignoreUnknownArguments ? doNothing : unknownOptionError;
+
+            int consumeNext = 0;
+            Action<int> onConsumeNext = (n => consumeNext = consumeNext + n);
+            bool forceValues = false;
+
+            var tokens = new List<Token>();
+
+            var enumerator = arguments.GetEnumerator();
+            while (enumerator.MoveNext())
+            {
+                switch (enumerator.Current) {
+                    case null:
+                        break;
+
+                    case string arg when forceValues:
+                        tokens.Add(Token.ValueForced(arg));
+                        break;
+
+                    case string arg when consumeNext > 0:
+                        tokens.Add(Token.Value(arg));
+                        consumeNext = consumeNext - 1;
+                        break;
+
+                    case "--" when allowDashDash:
+                        forceValues = true;
+                        break;
+
+                    case "--":
+                        tokens.Add(Token.Value("--"));
+                        if (posixlyCorrect) forceValues = true;
+                        break;
+
+                    case "-":
+                        // A single hyphen is always a value (it usually means "read from stdin" or "write to stdout")
+                        tokens.Add(Token.Value("-"));
+                        if (posixlyCorrect) forceValues = true;
+                        break;
+
+                    case string arg when arg.StartsWith("--"):
+                        tokens.AddRange(TokenizeLongName(arg, nameLookup, onBadFormatToken, onUnknownOption, onConsumeNext));
+                        break;
+
+                    case string arg when arg.StartsWith("-"):
+                        tokens.AddRange(TokenizeShortName(arg, nameLookup, onUnknownOption, onConsumeNext));
+                        break;
+
+                    case string arg:
+                        // If we get this far, it's a plain value
+                        tokens.Add(Token.Value(arg));
+                        if (posixlyCorrect) forceValues = true;
+                        break;
+                }
+            }
+
+            return Result.Succeed<IEnumerable<Token>, Error>(tokens.AsEnumerable(), errors.AsEnumerable());
+        }
+
+        public static Result<IEnumerable<Token>, Error> ExplodeOptionList(
+            Result<IEnumerable<Token>, Error> tokenizerResult,
+            Func<string, Maybe<char>> optionSequenceWithSeparatorLookup)
+        {
+            var tokens = tokenizerResult.SucceededWith().Memoize();
+
+            var replaces = tokens.Select((t, i) =>
+                optionSequenceWithSeparatorLookup(t.Text)
+                    .MapValueOrDefault(sep => Tuple.Create(i + 1, sep),
+                        Tuple.Create(-1, '\0'))).SkipWhile(x => x.Item1 < 0).Memoize();
+
+            var exploded = tokens.Select((t, i) =>
+                        replaces.FirstOrDefault(x => x.Item1 == i).ToMaybe()
+                            .MapValueOrDefault(r => t.Text.Split(r.Item2).Select(Token.Value),
+                                Enumerable.Empty<Token>().Concat(new[] { t })));
+
+            var flattened = exploded.SelectMany(x => x);
+
+            return Result.Succeed(flattened, tokenizerResult.SuccessMessages());
+        }
+
+        public static Func<
+                    IEnumerable<string>,
+                    IEnumerable<OptionSpecification>,
+                    Result<IEnumerable<Token>, Error>>
+            ConfigureTokenizer(
+                    StringComparer nameComparer,
+                    bool ignoreUnknownArguments,
+                    bool enableDashDash,
+                    bool posixlyCorrect)
+        {
+            return (arguments, optionSpecs) =>
+                {
+                    var tokens = GetoptTokenizer.Tokenize(arguments, name => NameLookup.Contains(name, optionSpecs, nameComparer), ignoreUnknownArguments, enableDashDash, posixlyCorrect);
+                    var explodedTokens = GetoptTokenizer.ExplodeOptionList(tokens, name => NameLookup.HavingSeparator(name, optionSpecs, nameComparer));
+                    return explodedTokens;
+                };
+        }
+
+        private static IEnumerable<Token> TokenizeShortName(
+            string arg,
+            Func<string, NameLookupResult> nameLookup,
+            Action<string> onUnknownOption,
+            Action<int> onConsumeNext)
+        {
+
+            // First option char that requires a value means we swallow the rest of the string as the value
+            // But if there is no rest of the string, then instead we swallow the next argument
+            string chars = arg.Substring(1);
+            int len = chars.Length;
+            if (len > 0 && Char.IsDigit(chars[0]))
+            {
+                // Assume it's a negative number
+                yield return Token.Value(arg);
+                yield break;
+            }
+            for (int i = 0; i < len; i++)
+            {
+                var s = new String(chars[i], 1);
+                switch(nameLookup(s))
+                {
+                    case NameLookupResult.OtherOptionFound:
+                        yield return Token.Name(s);
+
+                        if (i+1 < len)
+                        {
+                            // Rest of this is the value (e.g. "-sfoo" where "-s" is a string-consuming arg)
+                            yield return Token.Value(chars.Substring(i+1));
+                            yield break;
+                        }
+                        else
+                        {
+                            // Value is in next param (e.g., "-s foo")
+                            onConsumeNext(1);
+                        }
+                        break;
+
+                    case NameLookupResult.NoOptionFound:
+                        onUnknownOption(s);
+                        break;
+
+                    default:
+                        yield return Token.Name(s);
+                        break;
+                }
+            }
+        }
+
+        private static IEnumerable<Token> TokenizeLongName(
+            string arg,
+            Func<string, NameLookupResult> nameLookup,
+            Action<string> onBadFormatToken,
+            Action<string> onUnknownOption,
+            Action<int> onConsumeNext)
+        {
+            string[] parts = arg.Substring(2).Split(new char[] { '=' }, 2);
+            string name = parts[0];
+            string value = (parts.Length > 1) ? parts[1] : null;
+            // A parameter like "--stringvalue=" is acceptable, and makes stringvalue be the empty string
+            if (String.IsNullOrWhiteSpace(name) || name.Contains(" "))
+            {
+                onBadFormatToken(arg);
+                yield break;
+            }
+            switch(nameLookup(name))
+            {
+                case NameLookupResult.NoOptionFound:
+                    onUnknownOption(name);
+                    yield break;
+
+                case NameLookupResult.OtherOptionFound:
+                    yield return Token.Name(name);
+                    if (value == null) // NOT String.IsNullOrEmpty
+                    {
+                        onConsumeNext(1);
+                    }
+                    else
+                    {
+                        yield return Token.Value(value);
+                    }
+                    break;
+
+                default:
+                    yield return Token.Name(name);
+                    break;
+            }
+        }
+    }
+}
--- a/src/CommandLine/Core/GetoptTokenizer.cs
+++ b/src/CommandLine/Core/GetoptTokenizer.cs
@@ -88,36 +88,6 @@ namespace CommandLine.Core
             return Result.Succeed<IEnumerable<Token>, Error>(tokens.AsEnumerable(), errors.AsEnumerable());
         }
 
-        public static Result<IEnumerable<Token>, Error> ExplodeOptionList(
-            Result<IEnumerable<Token>, Error> tokenizerResult,
-            Func<string, Maybe<char>> optionSequenceWithSeparatorLookup)
-        {
-            var tokens = tokenizerResult.SucceededWith().Memoize();
-
-            var exploded = new List<Token>(tokens is ICollection<Token> coll ? coll.Count : tokens.Count());
-            var nothing = Maybe.Nothing<char>();  // Re-use same Nothing instance for efficiency
-            var separator = nothing;
-            foreach (var token in tokens) {
-                if (token.IsName()) {
-                    separator = optionSequenceWithSeparatorLookup(token.Text);
-                    exploded.Add(token);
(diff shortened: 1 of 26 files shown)


rmunn:
The failing unit tests are expected, and are fixed by #682. So #682 should be merged first, and then I'll rebase this PR on top of the new head of the `develop` branch so that the unit tests will pass.

rmunn:
Just rebased on top of the new state of `develop` now that #682 has been merged.

rmunn:
@moh-hassan - https://github.com/commandlineparser/commandline/pull/684/commits/47618e0ddf43a380baaa54484b188118b30d8a6c contains a fix for eight different bugs (listed in the PR description above). I've chosen to include it in this PR because GetoptTokenizer.cs (introduced in this PR) had the same bug, and needed the same fix, as the original Tokenizer.cs. (Because the buggy code is a method that I copied over from the original Tokenizer without modifying it). I thought about whether to submit that bugfix as a separate PR, but that would have caused *this* PR to get rather messy (this PR would then depend on another PR that had to be committed first, and it would have been rather complicated to get them merged in the right order, kind of like what happened with my original #607).

However, if you decide that you do **not** want to take this PR, please let me know and I'll do the work of separating out the Tokenizer bugfix from the rest of 47618e0ddf43a380baaa54484b188118b30d8a6c so that that bugfix, at least, can go in. But the simplest way by far is to just include that bugfix in this PR.

moh-hassan:
Thanks @rmunn  for this Great PR. 
I agree for the bugfix  included in this PR.
ParserSettings.GetoptMode =true/false (false is default) is used. 
To enable future extensibility and avoid breaking change, It's planned to support other Commandline standard like Posix and forward slash in windows and enable the corresponding tokenizer:

- I suggest use enum like:
```cs
enum ParserMode
{
Gnu ,  //default
GetOpt , 
//Posix,   
//SingleDash , //like powershell , Unity engine, msbuild
//ForwardSlash  // windows 
 
}
```
you can rename the options.

- Can you provide specialized test cases that are unique to GetOptMode?
- Can you provide a wiki page how to use GetoptMode with code example?
- It's nice to share the common code between Tokenizer and GetoptTokenizer for maintenance.

rmunn:
Hmm. The current parser mode isn't exactly GNU, because GNU getopt is what this PR is about emulating. I'm not quite sure what to call it, though: it's a weird mix of GNU behaviors and non-GNU behaviors. Perhaps "Default" mode would be a good name. And if there's ever a plan to change the default mode in the future (let's say there's widespread demand for Getopt mode to become the default in 3.x), then the current mode can be called "Legacy" mode when that breaking change is made, so that people who upgrade to 3.x but want to keep the 2.x behavior can just change "ParserMode = Default" to "ParserMode = Legacy" and the breaking change only forced them to change one line of code.

Or even better, we call the current mode "Legacy" right from the start, and *also* have an enum value called Default that is assigned to be the same as "Legacy" right now, but in 3.x might change to be assigned to some other mode. Yeah, I like that. That way people who use the default parser get the default mode, which might change in the future. And people who want to NOT have their current parser mode change in the future when we change the default can set it to "ParserMode = Legacy" right now, and we can guarantee that that won't change.

To answer your questions:

* This PR already includes specialized GetoptMode test cases, in the GetoptParserTests.cs file. Are they insufficient and you want to see more tests?
* I plan to work on the Wiki page next. I do want to wait until I know there will be no further changes needed to this PR and it will be merged as-is into the 2.9 release, because if (for example) I had already written up documentation for GetoptMode = true, I would have had to change it. So once you tell me, "Yes, this looks good, I plan to merge this without any more changes", I'll get started on the Wki documentation.
* The common code may not always stay common... but I guess if that happens we can copy the code from Tokenizer back into GetoptTokenizer and then make changes only to the GetoptTokenizer copy of the no-longer-common code. So I'll go ahead and deduplicate the common code now, and I can reverse that in the future if needed.

moh-hassan:
@rmunn 
The term `Legacy` may mean it's becoming obsolete and don't address that the library support GNU standard (with customization).
I'll revert the last commit and use Getopt =true/false.

moh-hassan (approved):

moh-hassan:
@rmunn 
Please, can you provide a wiki page for this feature with code example and guideline when to use getopt=true

rmunn:
@moh-hassan There were some documentation improvements in that commit as well as the change to using an enum for ParserMode, so I'll submit a new PR with just those documentation changes so that they don't get lost by your reverting commit b7102d8.

BTW, if you decided to revert b7102d8, does that mean that you're giving up on the idea of a ParserMode enum in the future that would allow SingleDash or ForwardSlash options? Because if we release 2.9 with GetoptMode as a boolean, that will mean that we can't easily change to the ParserMode system in the future without breaking backwards compatibility for everyone who was using GetoptMode = true in their code. So instead of just reverting b7102d8, I wish you had just edited it to change the name Legacy to something you liked better. I picked Legacy because I think people *will* want Getopt mode to become the default in the future, but you could have easily edited it to call it Gnu if that's your preference. I think reverting that commit was a mistake, but as long as we fix that mistake before pushing the 2.9 release out the door, it won't become TOO big a problem. If we push 2.9 out the door as it stands now, then you're going to have backwards compatibility problems when you finally create the ParserMode enum that you asked for in https://github.com/commandlineparser/commandline/pull/684#issuecomment-677881462.

rmunn:
@moh-hassan BTW, I'm going to start working on a Wiki page now.

moh-hassan:
@rmunn 
The commit [b7102d8](https://github.com/commandlineparser/commandline/commit/b7102d89a90acd2d9356637ead5a0a64b58b131a) can be restored with changing the options of enum 
Do you agree for the next names?
```
CustomGetOpt  //default
GetOpt
```

rmunn:
Yes, CustomGetopt would work. (The choices should be spelled Getopt and CustomGetopt, not GetOpt and CustomGetOpt, by the way: see [here](https://www.gnu.org/software/libc/manual/html_node/Getopt.html), for example, where the GNU documentation capitalizes the word getopt as Getopt, not as GetOpt).

rmunn:
@moh-hassan Forgot to tag you on my reply, so here's a tag so you get a notification. As I said, the choices should be spelled CustomGetopt and Getopt (with a lowercase o, not uppercase O) but apart from that I'm okay with the name CustomGetopt instead of Legacy. It will need documentation explaining the difference between Getopt and CustomGetopt, but that's precisely what my wiki page is going to be. I'll follow the steps outlined in https://github.com/commandlineparser/commandline/issues/486 and submit patches. (Though if I'm going to be doing this a lot, I may end up asking for write access to the repo so that I can edit the wiki more directly and not have to bug you every time. But for now, I'll just submit a PR with my wiki edits).

rmunn:
@moh-hassan - I have an even better idea. Instead of calling the current mode CustomGetopt or Legacy, we should call it Classic mode. Unlike Legacy, that doesn't imply that the mode is old or deprecated. Rather, calling it Classic mode implies that: 1) this is the current behavior, and 2) this is a mode that will be supported long into the future *even if we change the default behavior at some point*. So instead of calling it CustomGetopt, when you put commit b7102d8 back in, please call it Classic mode so that there will be less confusion in the long run.

rmunn:
@moh-hassan I've opened https://github.com/commandlineparser/commandline/pull/690 so that you can just merge that in easily if you are happy with the name Classic.

moh-hassan:
@rmunn 
Thanks for the suggestion.
It's better for the name to express the standard "Getopt" it implement, so developer don't confuse when set Parsermode to Classic. 
CustomGetopt means Tokenize with customization: e.g. 'AllowDashDash', 'IgnoreUnknowArguments', ...
So,kindly,  I'll merge the PR #690 with the name "CustomGetopt"  (with small letter o :)".

rmunn:
@moh-hassan *Sigh*. That's a very bad decision that will just cause lots and lots of user confusion, but I can't stop you. The Getopt mode also allows the same amount of customization (you can turn AllowDashDash off if you want, IgnoreUnknownArguments works as designed, and so on), so haivng a CustomGetopt mode and a Getopt mode is going to do nothing but confuse people. I'm working on Wiki documentation explaining in detail the difference between Classic mode and Getopt mode, and I can certainly rename Classic to CustomGetopt in that documentation if you do carry through with that bad decision. But I again **URGE** you to reconsider and not make the mistake of calling these two modes something so similar that people will constantly get them confused. That's a mistake that will haunt the library forever if you make it now. Please don't do it.

P.S. Earlier I was okay with the name CustomGetopt, but that was before I started trying to explain the difference in documentation. Once I started to try to explain the difference I realized that the name CustomGetopt tells the user nothing about how it differs from Getopt, and so the similarity in the names is just going to confuse people for no reason. That's when I decided to call it Legacy, but Classic would have been even better which is why I went with Classic for #690.

moh-hassan:
@rmunn 

What about these options:
```
GetoptParserV1   //default
GetoptParserV2
```
Both implement Getopt and no confusion .

derrickstolee:
Hi, @rmunn. I'm interested in consuming a version of your package with this fix in https://github.com/microsoft/scalar. Do you have a timeline on the 2.9 release?

rmunn:
@moh-hassan I have finally found time to work on CommandLineParser again. I like the GetoptParserV1 and V2 names you suggested in https://github.com/commandlineparser/commandline/pull/684#issuecomment-682600499, and I've implemented that change in #690. I'm also working on a wiki page explaining the difference between GetoptParserV1 and V2, and I'll submit it soon. Meanwhile, I think it might be a good idea to merge #690 before the 2.9 release goes out, so that the parser mode can be an enum instead of boolean GetoptMode value — this will allow for future expansion in the parser modes, such as the "use / as the option character" mode that people from Windows backgrounds often ask for.

rmunn:
@derrickstolee The 2.9 release was held up by the old NuGet API key for the package having expired. #733 updates the NuGet API key, so once that gets merged the release will be possible. Watch #733 if you want to find out when that happens. Unfortunately, I can't do anything about the release timing, as I'm not one of the project maintainers. I believe @ericnewton76 is the one who manages releases right now; Eric, please correct me if that's mistaken.

rmunn:
#735 and #709 are also fixed by this PR, and should be closed once 2.9.0 is finally released.

PR #690, opened 2020-08-27
ParserMode setting as enum rather than GetoptMode as bool

rmunn:
This is https://github.com/commandlineparser/commandline/commit/b7102d89a90acd2d9356637ead5a0a64b58b131a with the Legacy name renamed to Classic so as not to suggest that it's obsolete or going away anytime soon. See https://github.com/commandlineparser/commandline/pull/684#issuecomment-681783784 for the rationale.

### DIFF
--- a/src/CommandLine/Core/GetoptTokenizer.cs
+++ b/src/CommandLine/Core/GetoptTokenizer.cs
@@ -88,36 +88,6 @@ namespace CommandLine.Core
             return Result.Succeed<IEnumerable<Token>, Error>(tokens.AsEnumerable(), errors.AsEnumerable());
         }
 
-        public static Result<IEnumerable<Token>, Error> ExplodeOptionList(
-            Result<IEnumerable<Token>, Error> tokenizerResult,
-            Func<string, Maybe<char>> optionSequenceWithSeparatorLookup)
-        {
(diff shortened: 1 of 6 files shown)


rmunn:
Enum names are now GetoptModeV1 and GetoptModeV2 as per discussion with @moh-hassan (see https://github.com/commandlineparser/commandline/pull/684#issuecomment-682600499 and previous comments for background).

rmunn:
@ericnewton76 - Now that there's some motion on the CommandLineParser project again, I'd like to ask you to take a look at this PR. Having `GetOptMode` as a boolean limits future design choices, and I think it should be an enum instead. Rationale in https://github.com/commandlineparser/commandline/pull/684#issuecomment-681783784. In https://github.com/commandlineparser/commandline/pull/684#issuecomment-682332784, @moh-hassan expressed an intention to merge this PR with the name "CustomGetopt"; I disagreed with the name, and as you can see, this PR was never merged.

At this point, changing the bool to an enum might be a breaking change, but it's a breaking change in a feature that was only just released so right now would be the time to make it. The longer that change waits, the more people will be affected by it. So please take a look at this PR (https://github.com/commandlineparser/commandline/pull/690) as soon as you can. Thanks.

psfinaki:
Anyone here? Please? @moh-hassan @ericnewton76?

PR #694, opened 2020-09-15
Set to true nullable bool arguments without the value

### LINKED ISSUE #316: Set to true nullable bool arguments without the value
Consider the following class:

```csharp
class Parameters
{
    [Option("bool")]
    public bool Bool { get; set; }
    
    [Option("nullable-bool")]
    public bool? NullableBool { get; set; }
}
```

It works fine for such arguments: `--bool --nullable-bool true` but doesn't work for `--bool --nullable-bool`.

I suggest making it working for nullable bool types.

Kira-NT:
Well, it fixes #316, what can I say ¯\\_(ツ)_/¯

Quick example:

```c#
class Parameters
{
    [Option("bool")]
    public bool Bool { get; set; }
    
    [Option("nullable-bool")]
    public bool? NullableBool { get; set; }
}
```

| --bool | --nullable-bool | Parameters(Bool, NullableBool) |
| ------- | ---------------- | ------- |
| + | `true` | (`true`, `true`) |
| + | `false` | (`true`, `false`) |
| + | + | (`true`, `true`) |
| + | - | (`true`, `null`) |
| - | `true` | (`false`, `true`) |
| - | `false` | (`false`, `false`) |
| - | + | (`false`, `true`) |
| - | - | (`false`, `null`) |

### DIFF
--- a/src/CommandLine/Core/KeyValuePairHelper.cs
+++ b/src/CommandLine/Core/KeyValuePairHelper.cs
@@ -15,6 +15,14 @@ namespace CommandLine.Core
             return tokens.Select(t => t.Text.ToKeyValuePair("true"));
         }
 
+        public static IEnumerable<KeyValuePair<string, IEnumerable<string>>> ForScalarSwitch(
+            IEnumerable<Token> tokens)
+        {
+            return tokens
+                .Group(2)
+                .Select((g) => g[0].Text.ToKeyValuePair(g[1].Text));
+        }
+
         public static IEnumerable<KeyValuePair<string, IEnumerable<string>>> ForScalar(
             IEnumerable<Token> tokens)
         {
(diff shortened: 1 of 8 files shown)


Kira-NT:
Welp, the repo is dead, so I think it's simpler and better to switch to [dotnet/command-line-api](https://github.com/dotnet/command-line-api)

detoxhby:
This should've been merged... essential feature. Did all maintainers left the project? (If so, it's sad that such a handy project gone waste...)

Kira-NT:
@detoxhby, just switch to `System.CommandLine`. It is better and is maintained by dotnet community, so I don't see a chance for `dotnet/command-line-api` to die in the nearest future, just how this library did

kyleshrader:
Why was this never merged?

PR #698, opened 2020-10-05
Add environment variable support

### LINKED ISSUE #677: Can I read options from Environment variables?
First of all thanks for creating this nice package. Been using it for a lot of projects and it's working fine as always :) 

I was wondering if it's possible to read options from environment variables as well as from command params?
When I'm running my app in a docker container I would like to be able to edit my env variables in the container and then reflect that to my startup options. 

In the old repo, I found this issue https://github.com/gsscoder/commandline/issues/58 but can't seem to find the implementation in the new repo. Was it never implemented or planned as a feature or are they other ways to resolve my "problem"?

If not I can always just read the env variables myself and update the startup options if needed.

jannickj:
Fixes: https://github.com/commandlineparser/commandline/issues/677

Allows for enviroment variables to be part of a cli setup the order is Args > Env > Default
NB. Currently does not support sequences (Future work - will instead crash with a helpful error msg)

I also changed 0/1 to be parsable booleans as that's the language agnotisc / most common way to set booleans via environment variables.

### DIFF
--- a/src/CommandLine/BaseAttribute.cs
+++ b/src/CommandLine/BaseAttribute.cs
@@ -12,6 +12,7 @@ namespace CommandLine
         private int min;
         private int max;
         private object @default;
+        private string env;
         private Infrastructure.LocalizableAttributeProperty helpText;
         private string metaValue;
         private Type resourceType;
@@ -84,6 +85,18 @@ namespace CommandLine
             set
             {
                 @default = value;
+            }
+        }
+
+        /// <summary>
+        /// Gets or sets mapped property env value.
+        /// </summary>
+        public string Env
+        {
+            get { return env; }
+            set
+            {
+                env = value;
             }
         }
 
--- a/src/CommandLine/Core/InstanceBuilder.cs
+++ b/src/CommandLine/Core/InstanceBuilder.cs
@@ -88,7 +88,7 @@ namespace CommandLine.Core
                 T instance;
                 if(typeInfo.IsMutable() == true)
                 {
-                    instance = BuildMutable(factory, specPropsWithValue, setPropertyErrors);
+                    instance = BuildMutable(factory, specPropsWithValue, setPropertyErrors, ignoreValueCase, parsingCulture);
                 }
                 else
                 {
@@ -126,7 +126,12 @@ namespace CommandLine.Core
             return result;
         }
 
-        private static T BuildMutable<T>(Maybe<Func<T>> factory, IEnumerable<SpecificationProperty> specPropsWithValue, List<Error> setPropertyErrors )
+        private static T BuildMutable<T>(
+            Maybe<Func<T>> factory,
+            IEnumerable<SpecificationProperty> specPropsWithValue,
+            List<Error> setPropertyErrors,
+            bool ignoreValueCase,
+            CultureInfo parsingCulture)
         {
             var mutable = factory.MapValueOrDefault(f => f(), () => Activator.CreateInstance<T>());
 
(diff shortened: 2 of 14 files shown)


jannickj:
Hello?

jannickj:
:wave:

jannickj:
:cloud_with_rain: 
 :walking_man:

SommerEngineering:
@ericnewton76 Could this be part of the 2.9.0 release, please? 🙂

jannickj:
:sleeping_bed:

jannickj:
:sun_behind_small_cloud:

jannickj:
:sun_with_face:

BoBiene:
@ericnewton76 & @moh-hassan would be great to get this feature :)

johnjaylward:
It looks like @moh-hassan has left this project (he's not listed in the org anymore), and @ericnewton76 has been inactive on GitHub since November. It effectively looks like this project has been abandoned, which is a shame :(

BoBiene:
So there is no one able to merge pullrequests to this repo?
@gsscoder you are listed as contact, are you able to maintain this great project?

jannickj:
:city_sunrise:

jannickj:
:cityscape:

jannickj:
:cricket:

jannickj:
:skull:

devedse:
I created a NuGet package based on your branch:
https://www.nuget.org/packages/CommandLineParserWithEnvironmentVariables/2.8.0

devedse:
@jannickj , I did find one issue though. If `Required = true` and you did provide the value for the variable through an EnvironmentVariable it's still showing the error that a required-option is missing.

Not sure if that is desirable.

jannickj:
@devedse thx :D, also would say that is in my view expected behavior :)

devedse:
@jannickj , the thing is that this results in me having to check all variables in the code to see if they are `not null`. Even though it is required, thus, should be provided.

But I'm not sure what the best solution would be.

jannickj:
yeah it's a bit annoying the way to get around it is to just set it equal to `null!`
ex.
```csharp

[Option("host",
	HelpText = "The host for the client",
	Default = "localhost:80")]
public string Host { get; set; } = null!;
```

devedse:
That's not really what I mean. What I mean is that if you set Required = true, you need to provide the variable by passing in the flag.

For example, let's assume we have:

```
[Option("data",
        Required = true,
	HelpText = "The data",
        Env = "DATA"
)]
public string Data { get; set; } = null!;
```

```
#This would work:
testapp --data 123
```

```
#This would not work:
SET DATA=123
testapp

#Now we're getting the message that --data is missing.
```

jannickj:
right... I wasn't clear in my first message then, I tried to argue that it could make sense since required sort of strongly requests that you set the field. That said I understand what you mean and the more I think about the more I'm convinced that it's the behavior you expect.

devedse:
That's what I thought first as well. But I'm doubting it a bit now since if you look at it from the consumer end it just requires the data. It doesn't matter where it comes from. I can't think of a scenario where actually passing it as a flag is required if there is also an environment variable.

But either way is fine. I'll just work with the way is implemented now.

jannickj:
> That said I understand what you mean and the more I think about the more I'm convinced that it's the behavior you expect.

just read what I wrote, what I meant to say is that the behavior you (@devedse ) expected is the correct behavior, i.e. required accepts environment variables :sweat_smile:

devedse:
@jannickj , are you planning on implementing this?

jannickj:
@devedse probably not, I made this PR 1.5 years ago, so at this point I'm okay with it, if someone revived this git-repo then maybe but not on a half-dead project.

devedse:
I'll see if I can make a quick fix then. I'll just release a new version of my NuGet Package

jannickj:
:muscle:

devedse:
I think I fixed it. I actually moved all the code for setting the Value from the `BuildMutable` Method to the `OptionMapper.MapValues` as this is probably the place where we'd want to add this stuff.
https://github.com/devedse/commandline/commit/9ef1591563d24d981352b0df089277f839fba64b

I also wrote some test scenario's that test `Required = true` in combination with providing the value through Environment values or arguments or neither.

I've released this version to NuGet:
https://www.nuget.org/packages/CommandLineParserWithEnvironmentVariables/2.8.1

devedse:
Feel free to do a code review, Incase I missed something.

Issue #109, opened 2017-11-04
Incorrent argument checks in HeadingInfo.cs


ericnewton76:
<a href="https://github.com/zii-dmg"><img src="https://avatars2.githubusercontent.com/u/2846896?v=4" align="left" width="96" height="96" hspace="10"></img></a> **Issue by [zii-dmg](https://github.com/zii-dmg)**
_Friday Jan 20, 2017 at 15:30 GMT_
_Originally opened as https://github.com/gsscoder/commandline/issues/408_

----

[Constructor](https://github.com/gsscoder/commandline/blob/master/src/CommandLine/Text/HeadingInfo.cs#L30):
```
public HeadingInfo(string programName, string version = null)
{
    if (string.IsNullOrWhiteSpace("programName")) throw new ArgumentException("programName");
```
Passing string `"programName"` instead of argument `programName` value. But HeadingInfo.Empty is creating object with empty string! So I suppose this check should removed and xml comment with exception description.

[WriteMessage](https://github.com/gsscoder/commandline/blob/master/src/CommandLine/Text/HeadingInfo.cs#L104):
```
if (string.IsNullOrWhiteSpace("message")) throw new ArgumentException("message");
```
Same - `"message"` vs `message`.

PR #713, opened 2020-11-18
fixed string.IsNullOrWhiteSpace() check

0x6d61726b:
Coincidentially I found and fixed an issue when checking ```string.IsNullOrWhiteSpace()```.


0x6d61726b:
This fix seems also be related to #109.

PR #726, opened 2020-12-14
Fixes PlatformNotSupported on Xamarin.Mac modern.

### LINKED ISSUE #725: System.Console.IsOutputRedirected Not supported on Xamarin.mac
Using commandlineparser on Xamarin.Mac Modern throws a PlatformNotSupportedException when trying to set MaximumDisplayWidth. I found some older issues with docker where setting `MaximumDisplayWdith` explicitly would fix the issue. Unfortunately this won't work here, as the call to `System.Console.IsOutputRedirected` happens in the ParserSettings constructor.

mmoraga:
On Xamarin.Mac `Console.IsOutputRedirected` throws a `PlatformNotSupportedException`.
This PR moves that line to within the try-catch-block so it can be safely ignored. 
Fixes #725

### DIFF
--- a/src/CommandLine/ParserSettings.cs
+++ b/src/CommandLine/ParserSettings.cs
@@ -41,13 +41,12 @@ namespace CommandLine
 
         private int GetWindowWidth()
         {
-
-#if !NET40
-            if (Console.IsOutputRedirected) return DefaultMaximumLength;
-#endif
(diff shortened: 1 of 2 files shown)


mmoraga:
Fixes: https://github.com/commandlineparser/commandline/issues/725

PR #727, opened 2020-12-14
Better error message for missing value arguments

### LINKED ISSUE #363: Custom error message for missing sequence values
Hello, thanks for this great library.

I have a question regarding sequence values. I've set a property for sequence of values like this:
```csharp
        [Value(0, Min = 1, HelpText = "Path(s) to file(s)", Hidden = true)]
        public IEnumerable<string> FilePaths { get; }
```
So at least one value is required. When I parse an empty list of arguments the generated help message contains this:
```
ERROR(S):
A sequence value not bound to option name is defined with few items than required.
```
That message is not very informative to the user (it also contains a grammatical error if I'm not mistaken: `few` should be `fewer`?). How can I customize it? Thanks in advance.

hol430:
Fixes #363

The problem shows up when using an argument like this:

```csharp
[Value(0, Min = 1, HelpText = "Path(s) to file(s)", Required = true)]
public IEnumerable<string> FilePaths { get; }
```

Calling the CLI without providing a value for this argument will
currently result in the following error:

> ERROR(S):
> A sequence value not bound to option name is defined with few items than required.

This patch changes the behaviour so that, iff a MetaName is provided in the ValueAttribute, the user of the CLI will receive a message along the lines of:

> ERROR(S):
>   Required option '$MetaName' is missing.

### DIFF
--- a/src/CommandLine/Core/NameExtensions.cs
+++ b/src/CommandLine/Core/NameExtensions.cs
@@ -20,12 +20,23 @@ namespace CommandLine.Core
                 specification.LongName);
         }
 
+        public static NameInfo FromValueSpecification(this ValueSpecification specification)
+        {
+            return new NameInfo(
+                specification.MetaName,
+                specification.MetaName);
+        }
+
         public static NameInfo FromSpecification(this Specification specification)
         {
             switch (specification.Tag)
             {
                 case SpecificationType.Option:
                     return FromOptionSpecification((OptionSpecification)specification);
+                case SpecificationType.Value:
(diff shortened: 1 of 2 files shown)


dalehenries:
Any chance this could be merged?  It looks like this project isn't very active.  Is anyone maintaining this?

JohnGrat:
bump

hol430:
Ha, I had forgotten that this PR even existed. And the issue has been around for longer still :). @ericnewton76 - apologies for the ping, but is there anything I can do to help get this merged? I'm happy to tweak this as needed. Just let me know if there's anything you need from me.

PR #733, opened 2021-01-28
Updated NuGet API key

gsscoder:
As request. As in the subject.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -53,7 +53,7 @@ deploy:
 
 - provider: NuGet
   api_key:
-    secure: Ab4T/48EyIJhVrqkfKdUxmHUtseEVuXuyrGACxZ0KN35rb/BzABlBM2YjZojicvT
+    secure: oy2h2spmvhjrr6pnv7xae6jy4fdm3tzujpsgyxzt4dcmly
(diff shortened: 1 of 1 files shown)


rmunn:
@moh-hassan @ericnewton76 I'd like to get #690 merged before a release happens, so that the Wiki documentation I'm working on for the new GetoptModeV2 parser mode will be correct, and so that the new Getopt parser mode from #684 will be selected by an enum rather than a boolean (which will allow for future feature expansion better than a boolean GetoptMode switch would).

rmunn:
@ericnewton76 @moh-hassan Ping. Are you still getting notifications from this repo? I want to make sure #690 gets merged before the release, so that #684 is in the best shape it can be for 2.9.0. But since I'm not a maintainer, all I can do is keep pinging you until you respond. :-)

DmitriyBobrovskiy:
@rmunn looks like this library is not maintained anymore, is it?

rpavlik:
As a stop-gap measure, could packages be published to the GitHub Registry associated with this repo? https://docs.github.com/en/packages/working-with-a-github-packages-registry/working-with-the-nuget-registry

I can't figure out if there's an easy way to install from release assets otherwise.

PR #764, opened 2021-06-26
Remove System.Linq.Expressions dependency.

NN---:
The code is much simpler and it allows porting library to older framework.
https://github.com/theraot/Theraot/issues/157.


NN---:
@moh-hassan Any update ?

ericnewton76 (commented):
Is there a SystemLinqExpressions nuget package that could be removed too?

nvm I see whats going on

ericnewton76 (approved):
This looks like a good change, but I cant readily see the impact the change will make.  I'm a little worried about merging it but I think it'll work.

Issue #685, opened 2020-08-19
Parse Single dash option as dashdash option


Ducatel:
Hi everyone,

I'm trying to parse a command line (imposed, no way to change it) which are like
```
myexe.exe  -Url=https://blabla.com/ --AppName=val1 -e testCase  -monoapp Open -Ctx "default ctx" 
```

As you can see, there is a lot of different case:

 * Some single dash option with a long name with space as separator
 * Some single dash option with a long name with equals as separator
 * Single dash option with short name  with space as separator
 * .....

And I wasn't able to parse it because it seems that when there is just a dash option, it mustn't be a long name. ( I have a lot of `UnknownOptionError` which seem to come from that)

![image](https://user-images.githubusercontent.com/1184332/90609441-50b4e380-e204-11ea-9d14-e7b7de205cd3.png)

So I have a very simple code to test

```C#
[Verb("Open")]
public class CommandLineOptions
{
   [Option("Url", Separator = '=')]
   public string Url { get; set; }
   [Option("AppName", Separator = '=')]
   public string AppName { get; set; }
   [Option('e',"env", Separator = ' ')]
   public string Env { get; set; }
   [Option("monoapp")]
   public bool Monoapp { get; set; }
   [Option("Ctx", Separator = ' ')]
   public string Ctx { get; set; }
}

....... In another file but in the  same  assembly

static void Main(string[] args) 
{
        var commandLineOptions = new CommandLineOptionsOpen();
        var parser = new Parser(with =>
        {
            with.EnableDashDash = false; // tested as true, same result
            with.IgnoreUnknownArguments = false;
        });
        var parseResult = parser.ParseArguments<CommandLineOptionsOpen>(args).WithParsed<CommandLineOptionsOpen>(result => commandLineOptions = result);

        System.Console.WriteLine(parseResult.Tag); // NotParsed
        System.Console.WriteLine(commandLineOptions.Url); // Null like all others values

}

```

So I did something wrong ? There is a way to parse this command line ?

Thanks in advance for your help

rmunn:
The `EnableDashDash` option has nothing to do with allowing `-Url` to be treated as `--Url`. What it enables is the standard handling of a bare `--` arg (that is, an argument that's just two dashes and nothing else) as it's normally handled by the GNU [getopt_long() function](https://www.gnu.org/software/libc/manual/html_node/Getopt.html), where a bare `--` argument means "everything after this point is a value, not an option, even if it starts with a hyphen".

[...]

PR #767, opened 2021-08-03
Add setting to interpret options starting with a single dash as long named options

MichaeIDietrich:
This PR adds a setting to use single dashes for command line options.

As several others in issue #685 we wanted to switch to this library, since it is more robust than other libraries available. But due to backwards compatibility we needed to support single dashes instead of double dashes for options else we would break our customer's scripts.
So we forked this library and added an option for single dash support and now would like to bring this to the official repo.

There is a new setting `OptionsParseMode` in `ParserSettings`

With three modes:

* `Default` is identical to the current behavior

Options that start with a double dash must be defined using their full name. 
(e.g. `git rebase --interactive`)
Options that start with a single dash are interpreted as list of short named options. 
(e.g. `git clean -xdf`)

* `SingleOrDoubleDash` 

Options that start with a single or double dash are interpreted as short or full named option. 
(e.g. `git rebase --interactive` or `git rebase -interactive` or `git rebase --i` or `git rebase -i`, but `git clean -xdf` would no longer be possible)

* `SingleDashOnly`

Options that start with a single dash are interpreted as short or full named option. 
(e.g. `git rebase -interactive` or `git rebase -i` )
Options that start with a double dash are considered as an invalid input. 
(e.g. `git rebase --interactive` would no longer be possible)

It would be great if you could consider this feature, since there are several library consumers that are looking for this option.

### DIFF
--- a/src/CommandLine/Core/GetoptTokenizer.cs
+++ b/src/CommandLine/Core/GetoptTokenizer.cs
@@ -16,7 +16,13 @@ namespace CommandLine.Core
             IEnumerable<string> arguments,
             Func<string, NameLookupResult> nameLookup)
         {
-            return GetoptTokenizer.Tokenize(arguments, nameLookup, ignoreUnknownArguments:false, allowDashDash:true, posixlyCorrect:false);
+            return GetoptTokenizer.Tokenize(
+                arguments,
+                nameLookup,
+                ignoreUnknownArguments: false,
+                allowDashDash: true,
+                posixlyCorrect: false,
+                optionsParseMode: OptionsParseMode.Default);
         }
 
         public static Result<IEnumerable<Token>, Error> Tokenize(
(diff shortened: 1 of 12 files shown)


ericnewton76:
Thank you for submitting.

Your pull request looks comprehensive, and good job with adding the tests.  I believe this warrants a minor version bump to 2.10

I'm a little apprehensive about deviating from the original getopt specs but I do see the benefit of migration (ahem Nuget.exe) and similar.

MichaeIDietrich:
Oh great, it would be really nice if this could be merged, so there is no need for our dev team to maintain a local fork anymore.
I will have another look on this PR next week. If I remember correctly, there was bug we have already fixed in our local fork. 
If so, I'll update the PR and report back. 👍

MichaeIDietrich:
OK, I updated the PR. There were hard coded occurrences of `--help` and `--version` that I missed before. But now they are also handled correctly in respect to the used mode.

_By the way this new mode is almost in use for one year in our business application and we received no issues from our customers, so it is kind of tested externally already._ :)

So from my perspective this PR is ready to be merged, if you don't have any further suggestions.

josesimoes:
Any chance this can be reviewed and merged? It would be a very much welcome improvement!

daniel-flemming:
This PR does exactly what we want to achieve in our application. It would be great if single dashes would be an option in the CommandLine library. We really like the functionality and how this library provides an easy to use and easy to maintain command line support. Unfortunately we also have the hard requirement to use single dashes.

Merging this would be great :-)

HavlisDe:
This is exactli what we need. Please merge

Issue #403, opened 2019-01-25
Proposal for change: Remove raising exception when Copyright or company assembly attributes are null and set with default values


moh-hassan:
**Problem**
Test fail in Xunit or Nunit framework in net4x when Copyright or Company assembly attributes are null

**Problem Analysis**
The problem is existing since version 2.3 and now in v2.4.3 but the behavior is different based on the test project is net4x.
The test fail for the test cases that test --help or --version or even errors for bad options in the commandline
See this   [issue   ](https://github.com/commandlineparser/commandline/issues/354) and [issue ](https://github.com/commandlineparser/commandline/issues/389) 
and [this ](https://github.com/nunit/nunit/issues/3084) in NUnit project

These issues can't be detected by the current unit test project because unit test uses the method `SetAttributeOverride' and bypass the `EntryAssemply` 

The CommandLineParser in Production is running Ok when reading AssemblyAttributes copyright and company, but in Test project in Full Framework net4x or latter it cause the failure of test  when reading assembly attributes. 
The test is OK in netcoreapp2.x

In production the entryAssemply is the console project itself which have copyright and company assembly attributes
but in test , the executer is the testhost framework itself which has null values for copyright and company assemply attributes.

This because the CommandlineParser raise Exception when CopyRight/Company   are  null see [this](https://github.com/commandlineparser/commandline/blob/master/src/CommandLine/Text/CopyrightInfo.cs#L120)

**Behavior in v2.3**
in net45x or latter: Assembly.GetExecutingAssembly();
in netstandard1.5 or latter: Assembly.GetEntryAssembly()

		 private static Assembly GetExecutingOrEntryAssembly()
				{
					var assembly = Assembly.GetEntryAssembly();

		#if !NETSTANDARD1_5
					assembly = assembly ?? Assembly.GetExecutingAssembly();
		#endif

					return assembly;
				}

**Behavior in v2.4.3**
in netstandrad2.x :Assembly.GetEntryAssembly();  

		 private static Assembly GetExecutingOrEntryAssembly()
				{
					return Assembly.GetEntryAssembly();
		}
						
**Advantage of removing raising exception** 

- The unit test don't fail
- No side effect at all for the flow of the CommandLineParser.
- Even, in test, the values of copyright and company is not valid in unit test and they are the values in hosttest assemblies and currently CommandlineParser pass artificial values for them in unit test.
- These values are different based on the version of testhost used and causes troubles to the unit test.
- Setting default values is better than firing exception.

[...]

PR #784, opened 2021-10-18
Allow explicit definition of "entry" assembly

mnivet:
This PR is related to my comment in #403 

I've not add this as a parser setting like I've suggest because it would have need more changes to make that non static.
Furthermore the need can be seen as an advanced feature needed only in some unit tests.

This both reasons lead me to this design that allow consumer to setup AutoBuildMetadataAssembly in its production code, or only in its test code.


mnivet:
I've just read about **develop** branch, so I close this PR and reopen one that can target this branch.

PR #807, opened 2022-03-09
Customizable type convesions

Ttxman:
I was searching for a way to use my own custom types to be parsed from arguments.
For extended set of "primitive" types, Commandlineparser already uses System.ComponentModel based type conversion.
So I just added this same conversion as fallback when creating by string only constructor fails.

With this you can decorate class with `TypeConverterAttribute` or register attribute programatically to type descriptor `TypeDescriptor.AddAttributes(.., new TypeConverterAttribute(...)));` and all other ways you can register conversion provider and you will get the custom types parsed from commandline attributes.


ericnewton76:
can you add any tests that are relevant?

Ttxman:
Sorry for the long delay, I just have too much work :/
I have added simulated tests around the usecase I needed: NET6 DateOnly type as commandline parameter.

Ttxman:
Dead thread and somewhat pointless implementation with the NET7+ IParsable static interfaces..

Issue #752, opened 2021-04-14
Constructor OptionAttribute(char, string) although property ShortName is string


HelOWeen:
The current constructor of ```OptionAttribute``` only allows for 1 character shortName, although the property ```ShortName``` itself is of type ```string```. Which prevents us from creating options with more than 1 character, though that might be useful in order to make the short name parameter name of similar options _(more)_ memorable for the user, e.g.

```
"fn", "filename"
"fs", "filesize"
```

PR #823, opened 2022-06-02
Fixes issue #752

vigneashs:
Current `OptionAttribute` have a limitation of shortName being able to be set to just one character. This creates a limitation in the functionality of giving more readable or rememberable short names. For example, -FileSystem can be made to -FS. 

This PR tends to address this limitation. 

PS: this my first PR with this repository, let me know if more test cases have to be added. 

Issue #752

### DIFF
--- a/src/CommandLine/OptionAttribute.cs
+++ b/src/CommandLine/OptionAttribute.cs
@@ -19,7 +19,13 @@ namespace CommandLine
         private char separator;
         private string group=string.Empty;
 
-        private OptionAttribute(string shortName, string longName) : base()
+
+        /// <summary>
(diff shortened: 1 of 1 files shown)


ericnewton76:
Thank you for submitting.

the design goal of shortname originates from the original getopt parser... you cant have a -FS if you have -F and a -S because how does the parser distinguish?  

if you want a short version of --FileSystem then create --FS option.

IMO I dont feel like this is a good move.

It could be argued to try to match Longest shortName first. but then there needs to be conflict reporting if a user creates 
```
[Option("FS", "FileSystem")]
public string FileSystem { get; set; }
//and
[Option("S", "FS")]
public string FileStatus { get; set; }
```

This adds a whole bunch of need for testing/conflict resolution and reporting logic to be added.

PR #831, opened 2022-06-19
Fix #830 - Recognize Option, Value and Usage attributes on private members

### LINKED ISSUE #830: Allow private property options
I think that private properties should be considered when reflecting to get all option properties.

This would allow for exposing a different interface for specifying options to the CLI and the code, which can often be beneficial.

It is already possible for the library to use a public property with a private `init`/`set` so this seems more like an oversight than anything else.

Saalvage:
### Potential use cases:
- Requiring additional parsing which requires type or multiplicity changes and wanting to avoid exposing those `init`/`set`-only properties unnecessarily, as usage of the Options class from a programming interface should not rely on the parsing and instead assign the parsed properties directly.
- The ability to encapsulate functionality in the Options class without having to expose the properties directly.
- Not wanting to expose the `Usage` functionality to the programming interface.

### Additional arguments:
- It's already possible to use private `init`/`set` on a public property, so this seems more like an oversight (or the other way around!)
- I don't think there are any good reasons to uphold this restriction.

### DIFF
--- a/src/CommandLine/Core/ReflectionExtensions.cs
+++ b/src/CommandLine/Core/ReflectionExtensions.cs
@@ -15,7 +15,10 @@ namespace CommandLine.Core
     {
         public static IEnumerable<T> GetSpecifications<T>(this Type type, Func<PropertyInfo, T> selector)
         {
-            return from pi in type.FlattenHierarchy().SelectMany(x => x.GetTypeInfo().GetProperties())
+            return from pi in type.FlattenHierarchy()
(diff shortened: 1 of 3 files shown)


ericnewton76:
The point of the Options class was only to hold the parser results.  Its a single-use design, that implies you have the ability to make public all options so the parser can apply parsed items from the commandline.

I'm a little dubious on making the change but since it doesnt actually hurt anything it'll probably go through

And BTW, thank you for also submitting the tests.

PR #832, opened 2022-06-22
Small cleanup

AntonC9018:
- Typo fixes;
- Make use of `nameof` for parameter names;
- Make use of pattern matching with `is` instead of `as` and null check.


ericnewton76:
Can you update/rebase your branch from current master?

AntonC9018:
not right now

PR #863, opened 2022-12-12
Fix #862. Properties with `required` modifier should be required.

### LINKED ISSUE #862: Treat .NET 7 `required` properties as actually required by default
**Code snippet:**
```cs
public class Options
{
    [Option]
    public required int Port { get; init; }

    [Option]
    public required string Server { get; init; }
}
```

**Expected behaviour:** Both `Port` and `Server` are required and the parsing fails if at least on of them is not present.

**Actual behaviour:** They are not required, I still have to specify `[Option(Required = true)]` in order for it to work

Orace:
Fix for #862

### DIFF
--- a/src/CommandLine/Core/Specification.cs
+++ b/src/CommandLine/Core/Specification.cs
@@ -4,6 +4,8 @@ using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Reflection;
+using System.Runtime.CompilerServices;
+using System.Runtime.InteropServices;
 using CommandLine.Infrastructure;
 using CSharpx;
 
@@ -54,7 +56,7 @@ namespace CommandLine.Core
             this.hidden = hidden;
         }
 
-        public SpecificationType Tag 
+        public SpecificationType Tag
         {
             get { return tag; }
         }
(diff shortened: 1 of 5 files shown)


tests/CommandLine.Tests/Unit/Issue862Tests.cs
```diff
@@ -0,0 +1,29 @@
...
+            var result = new Parser().ParseArguments<Options>(arguments);
+
+            result.Errors.Should().ContainSingle(e => e.Tag == ErrorType.MissingRequiredOptionError);
+            result.Tag.Should().Be(ParserResultType.NotParsed);
+        }
+
+        private class Options
+        {
+            [Option("cols")]
+            public required int Cols { get; set; }
```
DoctorKrolic, line 24: What if you have a `required` modifier and `[Option(Required = false)]`? I would say, it should throw because of conflicting settings. Please add test for such case

src/CommandLine/Core/Specification.cs
```diff
@@ -110,28 +112,40 @@ public bool Hidden
...
         public static Specification FromProperty(PropertyInfo property)
-        {       
+        {
             var attrs = property.GetCustomAttributes(true);
+
+#if NET7_0_OR_GREATER
+            var isRequired = attrs.OfType<RequiredMemberAttribute>().Any();
+#else
+            var isRequired = false;
+#endif
```
DoctorKrolic, line 122:
```suggestion
var isRequired = false;

#if NET7_0_OR_GREATER
            isRequired = attrs.OfType<RequiredMemberAttribute>().Any();
#endif
```

tests/CommandLine.Tests/Unit/Issue862Tests.cs
Orace: This is not a trivia 🤔
So I moved the discussion on #862 thread, [here](https://github.com/commandlineparser/commandline/issues/862#issuecomment-1347988738)

PR #865, opened 2022-12-12
Added `params` keyword for aliases in `VerbAttribute` constructor

### LINKED ISSUE #864: Cannot pass constant string array into VerbAttribute
The `VerbAttribute` constructor has the option to pass in `string[] aliases`, however attribute constructors should accept constants when used as attributes and, last I checked, arrays can't be constants.

I suggest using a `params` keyword.

rombethor:
Added `params` keyword to `VerbAttribute` constructor so that it can be used when forming the attribute as follows:

```c#
[Verb("my-command", false, "do", "the", "thing")]
public class MyCommandClass
{
   //does the thing
}
```

Closes #864


ericnewton76:
A lot of build errors in the test project related to this change.  Can we clean those up?

rombethor:
I think we can close this without a merge.  It was proven unnecessary in the issue #864

Issue #377, opened 2018-12-07
Example with Option with IEnumerable<string> broken - Sequence contains more than one element


ivberg-zz:
Running the latest commandlineparser v2.3.0 on .Net Core 2.1.

The library example doesn't seem to work at all parsing IEnumerable<string>
  [Option('r', "read", Required = true, HelpText = "Input files to be processed.")]
  public IEnumerable<string> InputFiles { get; set; }

If you specify --read file1 file2 you get this exception:

System.InvalidOperationException
  HResult=0x80131509
  Message=Sequence contains more than one element
  Source=System.Linq
  StackTrace:

>	System.Linq.dll!System.Linq.Enumerable.Single<System.Type>(System.Collections.Generic.IEnumerable<System.Type> source) Line 45	C#
 	CommandLine.dll!CommandLine.Core.InstanceBuilder.Build.AnonymousMethod__0_23(CommandLine.Core.SpecificationProperty sp)	Unknown
 	CommandLine.dll!CommandLine.Core.ReflectionExtensions.SetProperties.AnonymousMethod__0(System.__Canon current, CommandLine.Core.SpecificationProperty specProp)	Unknown
 	System.Linq.dll!System.Linq.Enumerable.Aggregate<CommandLine.Core.SpecificationProperty, LogMerger.Program.Options>(System.Collections.Generic.IEnumerable<CommandLine.Core.SpecificationProperty> source, LogMerger.Program.Options seed, System.Func<LogMerger.Program.Options, CommandLine.Core.SpecificationProperty, LogMerger.Program.Options> func) Line 55	C#
 	CommandLine.dll!CommandLine.Core.InstanceBuilder.Build.AnonymousMethod__16()	Unknown
 	CommandLine.dll!CommandLine.Core.InstanceBuilder.Build.AnonymousMethod__7()	Unknown
 	CommandLine.dll!CommandLine.Parser.ParseArguments<LogMerger.Program.Options>(System.Collections.Generic.IEnumerable<string> args)	Unknown

ivberg-zz:
Actually, this was an issue instead with Parsing a Dict
            public Dictionary<string, string> Dict { get; set; }

Is this supported?

nemec:
No, unfortunately dicts are not supported

ericnewton76:
A dictionary would be hard to support... what generates the keys for the values?

I would believe this is undefined in the getopt specs that this library strives to follow.

ivberg-zz:
[...]

PR #889, opened 2023-05-26
Throw invalid operation exception if Dictionary type is used for specification

phaniva:
To address [Issue 377](https://github.com/commandlineparser/commandline/issues/377), I have added to list of  existing specification gaurds.


PentaPinguino:
Hi Phani,

In my opinion the problem with this solution is that the error `Non scalar properties should be sequence of type IEnumerable<T>.` defined in `TypeConverter.ChangeTypeSequence` will be bypassed - that means that developers will receive a different error message when trying to parse a `IEnumerable` without a specified type.

In my opinion there are two possible solutions to the problem:

1. To ensure - inside `ChangeTypeSequence` - that the sequence has exactly one generic argument.
2. To add two new rules inside `SpecificationGuards` that ensure that sequence types have exactly one generic argument.

Is it fair to have some checks inside the `ChangeTypeSequence` and some checks inside `SpecificationGuards`? Shouldn't we maybe try to move all checks to `SpecificationGuards`?

Thanks in advance and kind regards,
Davide

PR #903, opened 2023-09-30
Fix ReadText.LocalizedDemo strings

mdanish-kh:
- If it's not a missing option value case, then the localized string should be `Properties.Resources.SentenceBadFormatConversionErrorValue` which is consistent with other cases in the code.
- Make `SentenceBadFormatConversionErrorOption` an english string


mdanish-kh:
Closing this and making a new PR targeting the `develop` branch as written in the contribution guide

PR #913, opened 2024-02-13
Make trimmable in net8.0

jjrdk:
I wanted to use the library in a command line tool and also make use of the trimming features. I added trimming attributes to make it work for .net 8.0. It will not work in any of the existing target frameworks for the project though.


kgamecarter:
Thx.
NativeAoT work fine.

olihou (approved):

